PR #151, opened 2025-07-20
Implementation of Dynamic String

### LINKED ISSUE #147: Replace char lexeme[500] with a dynamic string in lexer
Replace char lexeme[500] with a dynamic string in lexer:
Currently in the lexer we say that the maximum length of any lexeme is 500(499 + \0 at the end). While this is a good(if not too large) limit for identifiers, strings are often over this length. We should replace the lexeme field with a dynamically resizable string data structure.

jackr276:
Closes #147 

Dynamic strings are now in use for all lexer functions

All char str[] will be replaced with the dynamic string module for consistency, ease of use, and guaranteed dynamic resize-ability.


jackr276:
Ready to go, final confirmation and sign-off is needed.

jackr276:
All results confirmed, merge is good to go.

PR #152, opened 2025-07-21
Rewrite cfg pattern to use new result package strategy

### LINKED ISSUE #121: Speed up instruction selector, parser
The instruction selector and parser are currently our bottlenecks and they're written in a very ad-hoc way. Restructure these to enable faster processing.

### LINKED ISSUE #139: Complete CFG functionality rewrite
A complete rewrite of the CFG is needed

jackr276:
Closes #139 
Closes #121 

Rewrite cfg pattern to use new result package strategy. This avoids the need for any direct successor drilling.

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -217,7 +217,6 @@ dynamic_array_t* compute_reverse_post_order_traversal(basic_block_t* entry, u_in
 		//Go all the way to the bottom
 		while(entry->block_type != BLOCK_TYPE_FUNC_EXIT){
 			entry = entry->direct_successor;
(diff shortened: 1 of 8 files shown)


jackr276:
Special handling for emit_expression() needs to be done in preparation for ternary operations

jackr276:
Compound statement needs to be rewritten to use a switch/case strategy

jackr276:
Removal of statement in block would be helpful. It would also be helpful if the parser could validate break/continue statements instead of the CFG. Some kind of stack nesting would be useful, where we go through and check the nesting levels to see if a switch/loop is being held over a break.

jackr276:
All test results good, merge is approved.

PR #157, opened 2025-07-23
Ternary operation support

### LINKED ISSUE #136: Ternary Operations
Implement the ternary operator:

An example would be:
```
int x = y == 3 ? a else b;
```

### LINKED ISSUE #158: Ternary operations violate SSA rules
Ternary operations violate SSA rules. Temp variables are assigned to more than once.

jackr276:
DRAFT: Not ready

Complete support & test coverage for ternary operations in all scenarios

Closes #136 
Closes #158


jackr276:
Still need to do ternary inside of expression, which will be a more major effort as everywhere that invokes emit_expr will need to support block resizing. Unary expressions, primary expressions will also need to support this.

jackr276:
All pre-existing results have been confirmed, but ternary results have not yet been validated.

jackr276:
Looking at binary_operation_with_ternary, the temp var assignment violates SSA form. This will cause the optimizer to incorrectly optimize the expressions and get rid of too much. Needs to be addressed before merge.

jackr276:
We should experiment with not having the final temp assignment for the ternary - I don't know why we'd need it

jackr276:
Test results changed because of the way we now assign temp variables in binary operations after the right side emits. However, the ultimate assembly remains unchanged.

PR #163, opened 2025-07-29
Dual switch statement support

### LINKED ISSUE #153: Dual switch/case optionality
Add the ability to have two kinds of switch statements:

The one that we currently have no with no fall through and no ability to break out.

And a new one that is traditional C-style, where fall through does take place.

### LINKED ISSUE #164: Parser does not catch repeated default statements
Parser does not catch repeated default statements

If we have more than one default statement, we should have an error. The parser does not currently catch this.

### LINKED ISSUE #168: Fall through & return in case testing
Fall through & return in case testing

Adding an issue for this as it likely needs some more attention than the average test

jackr276:
Ollie language now supports two different kinds of switch statements

Also addresses issue where switch statements with more than one default in them are not caught by the parser.

Closes #153 
Closes #164 
Closes #168


jackr276:
Improvements may be possible, but this is looking good for now.

jackr276:
All checks good, will merge

PR #171, opened 2025-08-03
Function pointer support

### LINKED ISSUE #150: Function pointer implementation
Implementation of full support for function pointers in ollie language

### LINKED ISSUE #172: Function pointer type support in type system
Function pointer type support in type system.

Since we now have a new type to contend with, we'll need to ensure that it's fully supported throughout the type system

### LINKED ISSUE #173: Register allocator overruns bounds in "used_registers" array
The used register array is being overrun leading to memory corruption issues. This must be resolved urgently.

Image

### LINKED ISSUE #174: Array of function pointers
Make a test case for an array of function pointers and ensure that we support this

### LINKED ISSUE #175: Dynamic name construction for function pointer types
Dynamic name construction for function pointer types

Function pointer types should have their names dynamically created as their built, not done on command. We should do this using setter functions for the signatures as opposed to setting them by hand in the parser. Whenever we invoke the setter, we can dynamically create the name.

### LINKED ISSUE #176: Rewrite function call, function definition methods in parser to use function signatures
Rewrite function call method in parser to use function signatures.

When processing the parameters, we should be using function signatures instead of the parameter list in the symtab record. Since functions now all carry types with them, this should be an easy switch.

We should also be using it in the defining_previous_implicit area of the function_definition function:

Image

### LINKED ISSUE #177: Full support for function pointers, indirect function calls in CFG
Full support for function pointers, indirect function calls in CFG

Currently the CFG segfaults when it encounters a function identifier. We'll need to resolve this and add additional processing for the new AST_NODE_CLASS_INDIRECT_FUNCTION_CALL type that has been added by this change.

### LINKED ISSUE #178: Support for returning function types
It should be possible in ollie to have a function that returns a reference to a function type. We need to ensure that this is supported.

### LINKED ISSUE #179: Function pointer & identifier support in the optimizer
Full support for function pointers in mark & sweep, clean, etc. in the optimizer.

### LINKED ISSUE #180: Function pointer & identifier support in the instruction selector
Function pointer & identifier support in the instruction selector. This includes support in the simplifier and selector portions of the instruction selector.

### LINKED ISSUE #181: Register allocator support for function pointers
Register allocator support for function pointers.

Currently we get the following error: Fatal compiler error: variable found with that has no live range
rip_0

jackr276:
Full support for function pointer types in Ollie

Closes #150 
Closes #172 
Closes #173 
Closes #174
Closes #175
Closes #176
Closes #177
Closes #178
Closes #179
Closes #180 
Closes #181


oc/compiler/register_allocator/register_allocator.c
```diff
@@ -560,7 +560,7 @@ static void calculate_liveness_sets(cfg_t* cfg){
 			basic_block_t* func_entry = dynamic_array_get_at(cfg->function_entry_blocks, i);
 
 			//Reset the registers in here while we're at it
-			memset(func_entry->function_defined_in->used_registers, 0, sizeof(u_int8_t) * 17);
```
jackr276, line 563: #173

This small issue caused memory corruption, which lead to these hours of debugging

jackr276:
Final checks & sign off needed. Issues do still exist, but this basic functionality has been achieved. Any/all subsequent issues will be addressed separately.

PR #184, opened 2025-08-12
Returning in switch

### LINKED ISSUE #183: Function with a switch that returns in all paths segfaults
Function with a switch that returns in all paths segfaults

```
fn switcher(x:i32) -> i32{
switch(x){
case 1:
   ret 1;
case 2:
   ret 2;
default:
  ret 3;
}
}
```

This currently segfaults.

jackr276:
Returning in switch

Fixes to the way that returning through all control paths in a switch statement are handled.

Closes #183


jackr276:

PR #185, opened 2025-08-13
Indirect caller saving, improvments to register saving scheme

### LINKED ISSUE #182: Strategy needed for caller saving with indirect function calls
Strategy needed for caller saving with indirect function calls.

With an indirect function call, we don't know exactly what's being called. As such it's difficult to perform caller saving. We need a strategy that
at least mitigates this. This will be done separately from the main function pointer MR.

jackr276:
Indirect caller saving, improvments to register saving scheme

Closes #182


jackr276:
Callee-saving and stack allocation is still incorrect, the order is bad.

PR #191, opened 2025-08-16
Support for string constants

### LINKED ISSUE #129: String handling functionality
String handling functionality:

The compiler currently does not handle strings correctly at all. We need a standard way of doing this, from the lexer all the way to the final assembly.

### LINKED ISSUE #165: String initializers in Ollie
Support for string initializers in ollie language.

Depends on #129 : string handling functionality

### LINKED ISSUE #190: Support needed for "Local constant" blocks within functions
For string handling, we need support for local constant blocks that are generated whenever functions contain string constants.

Image

### LINKED ISSUE #192: Local constant support in optimizer
Local constant support in optimizer:

If we observe that a local constant is not at all used, we should not be printing it out in the function block headers. We should devise a strategy of doing this inside of the optimizer

jackr276:
Support for string constants like: char* = "my_string" in the ollie compiler

Closes #129 
Closes #165 
Closes #190 
Closes #192

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -374,6 +374,9 @@ static void print_block_three_addr_code(basic_block_t* block, emit_dominance_fro
 
 	//Print the block's ID or the function name
 	if(block->block_type == BLOCK_TYPE_FUNC_ENTRY){
+		//Print out any/all local constants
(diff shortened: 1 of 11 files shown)


jackr276:
All string-free files have no change in output

jackr276:
All results good, merge is approved.

PR #194, opened 2025-08-16
Main argument enforcement, public/private function support

### LINKED ISSUE #189: Enforcement of argument order in main()
In the main function, argc must come first and must be an i32. It must then be followed by a pointer to an array of strings called argv. Anything else for a function named main will not work and should cause an error.

### LINKED ISSUE #193: Public and private functions
Public and private functions:

"public" keyword exposes functions across all modules (.global) prefix
"private" keyword acts like static in C. Those functions are not global

jackr276:
Main argument enforcement, public/private function support

Closes #189 
Closes #193


jackr276:
#189 needs test cases before we can call it good

#190 would probably benefit from a test case or two as well

jackr276:

PR #211, opened 2025-08-25
Use count integration, full validation of initializers

### LINKED ISSUE #205: Full validation needed for all initializer types
Full validation needed for all initializer types

### LINKED ISSUE #210: Full integration of "use_count" into cfg, optimizer
Full integration of "use_count" into cfg, optimizer

We need to integrate the "use_count" field to let us know if temporary variables can be wiped out. If their use count is above one, then clearly not. This will also need to be added into the "sweep()" function as the deletion of a statement that has a temporary variable would decrease its use count.

We can likely repurpose the "add_used_variable" function and move the temp check inside. If it isn't temp, add it to the block's used var storage and if not - increase its use count.

jackr276:
Use count integration, full validation of initializers

Closes #205 
Closes #210

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -757,6 +757,17 @@ void delete_statement(instruction_t* stmt){
 		previous->next_statement = next;
 		next->previous_statement = previous;
 	}
+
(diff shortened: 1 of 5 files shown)


jackr276:

PR #235, opened 2025-09-03
Rewrite of the Ollie Type System

### LINKED ISSUE #233: enum_member_list parser should work like the struct_member_list
enum_member_list parser should work like the struct_member_list

### LINKED ISSUE #234: Complete rethink of type system
It's looking more and more like the type system needs to be rethought, especially these so-called "internal_types". I think there is a way to rewrite the whole thing without any of them and only gain functionality.

jackr276:
A complete rewrite of the ollie type system that improves efficiency and correctness.

Closes #234 
Closes #233

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -3731,7 +3731,7 @@ static generic_ast_node_t* ternary_expression(FILE* fl, side_type_t side){
  *
  * BNF Rule: <construct-member> ::= {mut}? <identifier> : <type-specifier> 
  */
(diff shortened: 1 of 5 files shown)


jackr276:
We need test cases for user-defined enums with repeated values, mismatched user-defined and not user-defined, etc

jackr276:
We also need a test case for trying to mix and match user-defined and non-user-defined, and a use case where we go up to 16 bit enums.

jackr276:
Stack alignment will be done as part of a separate MR

jackr276:
All results good, ready to merge.

PR #242, opened 2025-09-07
Union type support in Ollie

### LINKED ISSUE #224: Union Type in Ollie
Union Type in Ollie

Add support for the union type in ollie

### LINKED ISSUE #241: Excessive use of the "ident()" rule throughout the parser
Excessive use of the "ident()" rule throughout the parser:

This function is used to check for identifiers. The problem is that it creates an AST node to do it. Since this is so often used, it is bound to waste memory. We need to replace this with direct use of the lexer where applicable to improve parser efficiency and reduce complexity

jackr276:
Union type support in Ollie

Closes #241 
Closes #224


jackr276:
We need test cases for enum definition parsing with missing semicolons & duplicated members to ensure that this is actually working

PR #255, opened 2025-09-14
Add support for forced double constants

### LINKED ISSUE #254: Forced double constants
Forced double constants:

In the same way we can use L or l to force an integer to be a long, we should be able to use D or d to force a float to be a double.

jackr276:
Add support for forced double constants

Closes #254


jackr276:
Ready to be checked

jackr276:
This is good to go

PR #265, opened 2025-09-18
Rework of structure pointers & access, and support for the new union pointer operator

### LINKED ISSUE #264: Union pointer access, struct pointer access revamp
Union pointer access:

In the same way that we have :: for struct pointer access, let's use => for union pointer access

We should also standardize: : and I think -> for struct pointers
For unions : . and => for pointers

jackr276:
Rework of structure pointers & access, and support for the new union pointer operator

Closes #264


jackr276:
This is looking good to me, ready to go

PR #269, opened 2025-09-21
Fixes for postfix & prefix operations

### LINKED ISSUE #247: Pre/post increment is not properly working in its current state
Validation of pre/post increment for struct/array types

jackr276:
Fixes for postfix & prefix operations

Closes #247

### DIFF
--- a/oc/test_files/postfix_testing.ol
+++ b/oc/test_files/postfix_testing.ol
@@ -1,0 +1,42 @@
+/**
+* Author: Jack Robbins
+* This file is intended to test structure intializers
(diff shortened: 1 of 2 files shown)


jackr276:
We should look into not emitting the assignment if it's a basic type(not in an array, etc.) basic types don't need the extra attention that complex types do.

jackr276:
Postfix is generating properly on the front end. However, on the back-end, we do end up with orphaned values like %edx here:

image


We need to also validate that this does not happen for prefix as well.

PR #276, opened 2025-09-24
Replace := with =

### LINKED ISSUE #273: Replace assignment operator(:=) with regular equals
Replace assignment operator(:=) with regular equals

We may in the future wish to use the assignment operator for something new. As of right now, just use regular equals.

jackr276:
Replace := with =: we may one day want to use the := operator for something different. We'll replace it now with regular = to simplify the syntax and typing experience.

Closes #273


jackr276:
Only change are in compiler warning outputs where := is now =, we are good to merge

PR #293, opened 2025-09-29
Address of postfix expressions like struct and union members, array values

### LINKED ISSUE #292: New strategy needed for handling memory address of union, struct and array members
New strategy needed for handling memory address of union, struct and array members

### LINKED ISSUE #294: Bad assignment check - Attempt to assign expression of type i32[]* to variable of type i32**
Attempt to assign expression of type i32[]* to variable of type i32**

This should work. It needs to be resolved in the type system.

jackr276:
Address of postfix expressions like struct and union members, array values

Closes #292 
Closes #294


jackr276:
All looks good, ready to go

PR #298, opened 2025-10-02
Testing & support added for recursive structure and union definitions

### LINKED ISSUE #290: Recursive struct/union member *pointer* definitions
We should be able to do the C-equivalent of this:

```c
struct my_struct {
	struct my_struct* a;
	char c;
	int x[6];
	double y;
};
```

We should not be able to do this:
```c
struct my_struct {
// THIS IS INCOMPLETE
	struct my_struct a;
	char c;
	int x[6];
	double y;
};
```

### LINKED ISSUE #299: Add "type_complete" field so that the compiler can detect when we're using unfinished types in a definition
Add "type_complete" field so that the compiler can detect when we're using unfinished types in a definition

### LINKED ISSUE #300: Variables may not be given incomplete types at time of definition
These types are incomplete and should generate errors:

```
declare mut x:i32[][3];

declare mut a:i32[3][];
```

jackr276:
Testing & support added for recursive structure and union definitions

Closes #290
Closes #299 
Closes #300

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -5311,7 +5311,7 @@ static generic_type_t* type_specifier(FILE* fl){
 
 		//If we're trying to create an array out of a type that is not yet fully
 		//defined, we also need to fail out
-		if(current_type_record->type->type_complete == FALSE){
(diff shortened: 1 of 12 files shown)


jackr276:
These new files are validated for this MR. The issue that has arisen for the recursive_struct_pointer_member will be triaged tomorrow morning.

PR #312, opened 2025-10-06
All memory region addresses are now stack(rsp) relative

### LINKED ISSUE #311: Rewrite initializer method in CFG to emit a memory address every time it runs
Rewrite initializer method in CFG to emit a memory address every time it runs:

Currently it emits one big memory address statement at the beginning, but that is not what we're after anymore

### LINKED ISSUE #314: Emit temp vars for base addresses always
Emit temp vars for base addresses always:

For initializers, we should stop having statements like:
arr_0 <- memory address of arr_0

Because in technicality this is not proper SSA form and it looks confusing.

### LINKED ISSUE #315: Incorrect movX instruction being selected in some instances
Incorrect movX instruction being selected in some instances:

We have noticed this in: postfix_testing.ol, prefix_testing.ol, and struct_initializer.ol

Image

jackr276:
All memory region addresses are now stack(rsp) relative

Closes #311
Closes #314
Closes #315

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -7757,7 +7757,10 @@ static cfg_result_package_t visit_let_statement(generic_ast_node_t* node, u_int8
 			add_variable_to_stack(&(current_function->data_area), assignee);
 
 			//Emit the statement here to get the base address
-			instruction_t* mem_addr = emit_memory_address_assignment(assignee, emit_var(node->variable));
+			instruction_t* mem_addr = emit_memory_address_assignment(emit_temp_var(node->inferred_type), assignee);
+
+			//For later reference, this is what we should be using
+			assignee = mem_addr->assignee;
 			
 			//Add it into the block
 			add_statement(current_block, mem_addr);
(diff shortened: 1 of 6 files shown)


jackr276:
Our approach for initializers will be as follows:

When we encounter a struct/array/string initializer, all that we do is begin incrementing an offset variable. This variable will be passed between rules, being incremented each time as appropriate, until the we hit the end of initializer chain. Once the end has been hit, we're able to emit the address offset calculation all in one go. This saves us *a lot* of instructions in the long run because we're not emitting tons of temp variables that do offset calculations. It will likely require a rewrite of the entire system, but these potential savings are worth it.

jackr276:
Bug noticed in postfix_testing.ol:
image

Need to investigate & resolve before merging

jackr276:
Ready to be looked at, now just needs validation

jackr276:
Validated:
address_of_2d_array.ol
array_access_deref.ol
array_initializer_2d.ol
array_initializer.ol
array_in_struct_intializer.ol
array_of_structs_initializer.ol
array_optimizer.ol
double_pointer_struct.ol
pointer_reassignment.ol
postfix_testing.ol
prefix_testing.ol
prepost_op_testing.ol
string_initializer2.ol
string_initializer.ol
struct_initializer.ol
struct_pointer_access.ol
struct_pointer_deref.ol
struct_pointers.ol
type_conversion_small_large.ol

All pre-existing files validated.

jackr276:
All ready to go

PR #333, opened 2025-10-13
Rewrite of basic blocks

### LINKED ISSUE #331: Basic blocks *must* obey the basic block property until post register-allocation
Basic blocks *must* obey the basic block property until post register-allocation

The issues that we've been facing stem from how our basic blocks do not obey the basic block property post the optimization step.

Basic Block: a maximal sequence of straight line code with one entry point and one exit that is either a branch or jump.

A perfect example is like this one but they're everywhere:

Image

This messes with the liveness analysis that we perform in the register allocator, so we cannot have this until after we've completed the register allocation

### LINKED ISSUE #332: Introduction of branch command in OIR
Introduction of branch command in OIR:

Instead of using conditional jumps paired with direct jumps, we should instead use the branch command in OIR so that it's easier for us to keep obeying the basic block property.

### LINKED ISSUE #334: Fully deprecate the use of "direct_successor" in the CFG front end
We should not be using this field anywhere before we order the blocks in the optimizer but for some reason it is still required

### LINKED ISSUE #335: Ensure all control flow statements can handle empty bodies
Ensure all control flow statements can handle empty bodies:

This includes while, for, do-while and if

### LINKED ISSUE #336: User defined jumps inside of a branch statement
In order to support the user-defined jump statements in ollie, we'll need to support them in a branch statement as well.

### LINKED ISSUE #338: Remove all extra "add_successor" calls in CFG
Remove all extra "add_successor" calls in CFG

The emit_Jump function now does this for us, so we should not see any double calls in this case.

### LINKED ISSUE #339: Removing successors/predecessors in optimizer
Removing successors/predecessors in optimizer after a sweep happens:

Currently this does not happen at all, which leaves me to question how this was ever correct. At the very least it needs an investigation

### LINKED ISSUE #340: Immediate Postdominator is broken
Immediate Postdominator is broken:

The algorithm that we are using appears to be incorrect.

### LINKED ISSUE #341: Optimizer overhaul to use branches
Optimizer overhaul to use branches

### LINKED ISSUE #345: reverse_post_order and reverse_post_order_reverse_cfg computation should be a guaranteed execution, at the time when everything else is computed
reverse_post_order and reverse_post_order_reverse_cfg computation should be a guaranteed execution, at the time when everything else is computed.

This is seriously buggy right now and needs to have some serious time spent on it.

### LINKED ISSUE #346: Short circuit with inverse jumps is still incorrect
Short circuit with inverse jumps is still incorrect

This needs to be checked

### LINKED ISSUE #347: PostProcess module that does some aspects of branch reduction, useless code elimination
PostProcess module that does some aspects of branch reduction, useless code elimination

### LINKED ISSUE #348: Move constants into a new utility area
Move constants into a new utility area

### LINKED ISSUE #349: Test cases for inverse and jumping required
Let's also have a test case for inverse jumping where one of the parameters is a function call

jackr276:
Rewrite of basic blocks - all basic blocks now will truly obey the basic block property until post register-allocation.

Closes #331 
Closes #332
Closes #334
Closes #335 
Closes #336 
Closes #338 
Closes #339
Closes #340
Closes #341
Closes #345
Closes #346 
Closes #347 
Closes #348 
Closes #349

### DIFF
--- a/oc/compiler/register_allocator/register_allocator.c
+++ b/oc/compiler/register_allocator/register_allocator.c
@@ -725,6 +725,9 @@ static void assign_live_range_to_source_variable(dynamic_array_t* live_ranges, b
 
 /**
  * Construct the live ranges appropriate for a phi function
+ *
+ * Note that the phi function does not count as an actual assignment, we'll just want
+ * to ensure that the live range is ready for us when we need it
  */
 static void construct_phi_function_live_range(dynamic_array_t* live_ranges, basic_block_t* basic_block, instruction_t* instruction){
 	//Let's see if we can find this
@@ -732,9 +735,6 @@ static void construct_phi_function_live_range(dynamic_array_t* live_ranges, basi
 
 	//Add this into the live range
 	add_variable_to_live_range(live_range, basic_block, instruction->assignee);
-
-	//This does count as an assignment
-	add_assigned_live_range(live_range, basic_block);
 }
 
 


jackr276:
If statements have been validated, all appears to be working after switching to branches

jackr276:
Files are identical after the change

jackr276:
Look at that immediate postdominator again, may be a way to do it without memory usage in the visited array

jackr276:
Open issues as of EOD 10/18:
Postprocessor needs to fully support branch_reduce
Short-circuit logic does not work for at least for loops, needs thorough review

Overall excellent progress made today.

jackr276:
Valgrind shows up clean, short circuiting appears to work

jackr276:
Results good(minus ternaries), valgrind is clean, ready to go.

PR #360, opened 2025-10-23
Precoloring Interference

### LINKED ISSUE #359: Fully separate pre-coloring run for live ranges
The crawl that does pre-coloring should not be attached to the liveness calculation - it should instead be done afterwards

### LINKED ISSUE #362: Stack pointer interference is not properly computed
We are not coalescing ranges with the stack pointer when we should be, because inherently the stack pointer does not interfere with
anything else.

### LINKED ISSUE #363: Coalescing with the stack pointer can cause stack pointer corruption
Coalescing with the stack pointer can cause stack pointer corruption:

There is the potential after coalescing with the stack pointer that some mutating expression gets pulled in to have the stack pointer. We'll need to account for this rare possibility.

### LINKED ISSUE #364: Precoloring spill cost analysis
Precoloring spill cost analysis: would it be more advantageous for us to spill the lower spill cost live range or the higher spill cost live range. In general the higher the spill cost, the longer this will take.

jackr276:
Ollie compiler can now detect and handle interference at the precoloring stage

Closes #359 
Closes #362
Closes #363
Closes #364

### DIFF
--- a/oc/test_files/ret_interference.ol
+++ b/oc/test_files/ret_interference.ol
@@ -10,7 +10,7 @@ pub fn main() -> i32 {
 
 	do {
 		b = a + 3;
-		c = c + b;
+		c = a + b;
 
 	} while(a < 9);
 


jackr276:
Going to stop here with this and go with the improvements that we've made already

jackr276:
Looks good except 2 minor bugs noticed that will be addressed separately. These bugs were not introduced by this MR.

PR #375, opened 2025-10-28
Complete memory model overhaul

### LINKED ISSUE #374: New, clearer memory model IR is needed that does not use indirection
New, clearer memory model IR is needed that does not use indirection

The indirection model that we currently have creates a mess for us. Really we should not use it. We should overhaul the IR to use all load and store operations instead of the (<var>) dereferencing

### LINKED ISSUE #376: Complete rewrite of postfix expression memory access handling
Complete rewrite of postfix expression memory access handling

Currently the base address is baked into our calculations. We should instead just be calculating the offset and then using our new load/store operations with the offset and base address. This is cleaner, simpler to analyze, and produces less overall instructions.

### LINKED ISSUE #379: Postfix expression address calculation should never actually use the base address
Postfix expression address calculation should never actually use the base address

Currently the base address is part of the calculations - in reality this should never be the case

### LINKED ISSUE #382: Stack variables should also use "memory address of"
Stack variables should also use "memory address of". They are also not tied into the SSA system with assignments.

### LINKED ISSUE #383: Rewrite middle end to support new memory model
Opening an issue for tracking. The middle end will need a slight rewrite for our new model.

### LINKED ISSUE #384: Rewrite instruction selector simplifier to support new memory model
Rewrite instruction selector simplifier to support new memory model. Opening this generic issue for tracking

### LINKED ISSUE #385: Load & store should just use the offset variable as their constant offset
Load & store should just use the offset variable as their constant offset

We should have this occupied from the start, so we can use op1_const to have actual store values.

### LINKED ISSUE #386: Instruction selector fixes for memory model overhaul
Instruction selector fixes for memory model overhaul - opening this generic issue for tracking

jackr276:
Complete memory model overhaul

Closes #374
Closes #376
Closes #379
Closes #382 
Closes #383
Closes #384 
Closes #385
Closes #386


jackr276:
Initializer appears to be working, now we'll need to do the postfix

jackr276:
TODO:

Optimize away this pattern:
image

Issue #357, opened 2025-10-22
Converting move operatioons can be optimized with other moves


jackr276:
Converting move operatioons can be optimized with other moves:

Currently, these 2 operations remain entirely separate:
movsx (t4), t5
movq t5, t6

We can apply all of our same optimizations here, and just make this:
movslq (t4),t6

This is made possible now by the specificity of the movslq. Movsx/movzx does not allow us to do this.

jackr276:
Make sure that we stress test the scenario of an 32 bit integer being moved to an unsigned 64 bit integer, as this does not require a converting move at all. This edge case must be accounted for.

PR #400, opened 2025-11-09
Converting moves are now fully expressed with types

### LINKED ISSUE #305: Fully write out converting move operations
Fully write out converting move operations

So far we use movsx and movzx. I would really rather we use values like movslq, movswb, etc.

jackr276:
Converting moves are now fully expressed with types

Closes #305


jackr276:
Results here are looking good. This MR only replaces the old movzx and movsx directives with type specific ones. There is another issue, #357, which will use this enhancement to make memory moves be done using one converting move instruction instead of two.

PR #404, opened 2025-11-11
Memory movement instructions can now leverage converting moves

### LINKED ISSUE #357: Converting move operatioons can be optimized with other moves

### LINKED ISSUE #405: Instruction selector/cfg needs to handle the case of unsigned quad word from/to 32 bit value conversion
Instruction selector/cfg needs to handle the case of unsigned quad word from/to conversion:

Case like:

Memory(32 bit) -> u64:

We'll need something like:
movl (mem), 32 bit(%ecx) - but going forward we just have the 64 bit version of this register

And the same story for moving into  memory.

This is necessary because there is no explicit instruction that allows us to move from 32 bits to 64 bits while 0-extending(unsigned).

We also need test coverage on both ends for this edge case.

### LINKED ISSUE #407: Successive loads/assignments should be condensed to one instruction
See this example from memory_converting_moves.ol:

Image

This should really just be a load into y_0. We need to explore why this is not happening and implement the fix.

### LINKED ISSUE #409: Store instructions can not leverage expanding moves
Store instructions can not leverage expanding moves in the same way loads can. Instead, we need to have an initial expanding move instead of an expanding move in the memory operation itself.

Example:
Wrong:
```
movslq %eax, (%rsp)
```
Right:
```
movslq %eax, %rdx
movq %rdx, (%rsp)
```

We'll be doing this in the instruction selector as opposed to the CFG method that we did before.

jackr276:
Memory movement instructions can now leverage converting moves

Closes #357 
Closes #405 
Closes #407 
Closes #409

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -3805,7 +3805,7 @@ static cfg_result_package_t emit_unary_operation(basic_block_t* basic_block, gen
 			 */
 			} else {
 				//If the side type here is right, we'll need a load instruction
-				instruction_t* load_instruction = emit_load_ir_code(emit_temp_var(indirect_version->type), indirect_version);
+				instruction_t* load_instruction = emit_load_ir_code(emit_temp_var(unary_expression_parent->inferred_type), indirect_version);
 
 				//The dereferenced variable has been used
 				add_used_variable(current_block, indirect_version);
--- a/oc/test_files/addition_with_memory_access.ol
+++ b/oc/test_files/addition_with_memory_access.ol
@@ -17,6 +17,12 @@ pub fn test_subtract(a:i64, ptr:i32*) -> i64 {
 	ret *ptr - a;
 }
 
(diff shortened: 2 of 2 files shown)


jackr276:
Validated:
addition_with_memory_access.ol
array_in_union.ol
array_of_strings_initializer.ol
basic_union_type.ol
construct_pointer_access.ol
lex_test11.ol
memory_converting_moves.ol
struct_in_union.ol
struct_pointer_access.ol
struct_pointer_deref.ol
type_conversion_small_large.ol
u64_memory_conversions.ol
union_in_struct.ol
union_pointer_access.ol
union_pointer.ol

jackr276:
Results looking good, good to merge.

PR #417, opened 2025-11-17
Back-end processing now happens on a per-function level

### LINKED ISSUE #408: Per-function CFGs for efficiency
As opposed to having one giant CFG that we need to manipulate for all operations(liveness, etc.), we should have per-function CFGs that allow us to operate on each independently.

This way if we, for example, need to spill during register allocation for one function, we do not need to redo the entire program, just that one function. This will give us more control and a boost in efficiency in larger, more complex programs.

### LINKED ISSUE #416: Revamp memory management for ast_nodes
Revamp memory management for ast_nodes: We should be using a dynamic array to store and then deallocate all of them. We should not be using the linked-list strategy.

jackr276:
**Back-end processing now happens on a per-function level**

Previously, we operated on the entire CFG for each pass of the simplifier, selector and register allocator. This is very inefficient because some functions may require little attention whilst others require multiple passes. Doing things like this allows us to go on a per-function level.

Closes #408 
Closes #416

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -378,8 +378,18 @@ compiler_test_non_timed: oc
 compiler_test_allocation_only: oc
 	@for input in $(inputs); do \
 		output=$$(echo $$input | sed 's|^$(TEST_FILE_DIR)|$(OUTPUTTED_ASSEMBLY_DIR)|' | sed 's|\.ol$$|.s|'); \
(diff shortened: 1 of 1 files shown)


jackr276:
No results here are changed from the original. We are good to merge.

PR #422, opened 2025-11-19
[Front End]: Fixes to mutability checking and correctness

### LINKED ISSUE #371: Uninitialized variables are not recognized in one certain case
They're not recognized when we have:

declare c:i32;
...
...
while() {
   c = c + 1; //Should fail, but doesn't
}

This needs investigation & resolution.

### LINKED ISSUE #418: Extensive checking/lockdown of the "mut" keyword & mutability
**Extensive checking/lockdown of the "mut" keyword & mutability**

We need to ensure that this keyword is actually meaningful. As part of this issue we need to see test cases that encompass every possible scenario with the mut keyword/mutability failure. This includes:
- Mutability with unary operations
- Pointers to mutable objects(mutable/nonmutable)
- Mutable/nonmutable arrays and union objects

And more. Any cases where we find issues will need to be addressed.

jackr276:
[Front End]: Fixes to mutability checking and correctness

Closes #371 
Closes #418

### DIFF
--- a/oc/test_files/short_circuit.ol
+++ b/oc/test_files/short_circuit.ol
@@ -7,7 +7,7 @@ fn tester(mut arg:i32) -> void{
 }
 
 //Dummy
-fn really_long_function(arg:i32) -> i32 {
+fn really_long_function(mut arg:i32) -> i32 {
 	arg++;
 	ret arg;
 }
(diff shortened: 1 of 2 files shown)


jackr276:
This MR is good, but I think we're going to take a new approach to mutability checking by using the type system itself and not relying on our old, clunky variable system.

PR #426, opened 2025-11-21
[Type System]: Mutability is now a feature of the type system, not the variable system

### LINKED ISSUE #421: Address of an immutable variable cannot be used as a mutable value
Currently the compiler allows this to go through:

```
pub fn mut_union(mut x:custom_union*) -> i32 {
	//We'll use the special union pointer accessor here
	x->x = 4;
	x->y = 2;

	ret x->x;
}


pub fn main() -> i32{
	declare my_union:custom_union;

	@mut_union(&my_union);
	
	//Read as char
	ret my_union.ch;
}
```

This should fail, because &my_union is the address of an immutable union variable. We need extra checking in the parser areas for function params and addresses

### LINKED ISSUE #423: Mutability in the type system
Mutability in the type system: it may be instead more worth it to build mutability directly into the type system itself instead of using variables. This is an idea that needs some exploration, but it could work well. Opening this issue as research.

### LINKED ISSUE #424: Integrate mutability away from variables and into the type system
Currently, mutability is associated with symtab_variable_record_t types in the variable subsystem. There are several drawbacks to this approach. These include but are not limited to:
   1.)  We lose mutability tracking whenever any kind of reference is involved. Taking the address of a variable completely disconnects the assignee from the variable
  2.) We need to do extra checking in the parser for variables when we assign or change. This is on top of the "is_operation_valid_for_type" checks that we already do. If we integrate into the type system, we no longer need to deal with this issue.
  3.) We are unable to track mutations across functions

All of these lead me to think that our best move is to completely remove mutability from the variable subsystem and instead add it into the type subsystem.

### LINKED ISSUE #425: All line_number elements should be 32 bits
All line_number elements should be 32 bits

This is all over the place across the project - some places it's 16, others it's 32. We need to standardize this to make them all 32 bit. 16 bits is not large enough to encompass all possible kinds of files.

### LINKED ISSUE #427: Full mutability overhaul in types_assignable
Full mutability overhaul in types_assignable:

This function needs to check for the special rules that assigning mutable memory regions to one another has.

### LINKED ISSUE #429: Functions can now return mutable types
Functions can now return mutable types:

fn my_func() -> mut i32*{
...
}

These need to now be supported in function definitions, pointers, etc.

### LINKED ISSUE #430: Mut is now a part of the type-specifier rule
For sanity reasons and logical code flow, we should move the mut keyword around to be a part of the type specifier itself:

```
let mut x:i32 ....
```

would be:

```
let x:mut i32 ...
```

### LINKED ISSUE #431: Do not use symtab scopes for union, struct members
Do not use symtab scopes for union, struct members:

There is no good reason to be creating entire sheafs for this. We should just maintain a local dynamic array like we already have and search through that.

### LINKED ISSUE #433: Edge case with "mutable" void types
Edge case with "mutable" void types:

The type system needs to have mutable void types in order to support something like mut void*, but we should never allow a user
to create a mut void return type or something of the sort. We'll need checks in the system to prevent this.

jackr276:
[Type System]: Mutability is now a feature of the type system, not the variable system

Closes #421
Closes #423
Closes #424
Closes #425
Closes #427
Closes #429 
Closes #430 
Closes #431 
Closes #433

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -8048,8 +8048,22 @@ static generic_ast_node_t* declare_statement(FILE* fl, u_int8_t is_global){
 	switch(declared_var->type_defined_as->type_class){
 		//These all require a stack allocation, so a node is required
 		case TYPE_CLASS_ARRAY:
-		case TYPE_CLASS_STRUCT:
 		case TYPE_CLASS_UNION:
+			/**
+			 * Special warning here - if the user is declaring an array or a union as
+			 * immutable, they actually never initialize these memory regions. We should
+			 * throw a warning up for this
+			 */
+			if(declared_var->type_defined_as->mutability == NOT_MUTABLE){
+				//Throw up the warning here
+				sprintf(info, "Type \"%s\" is immutable. If you declare variable \"%s\" with this type, you may never be able to initialize it",
+								declared_var->type_defined_as->type_name.string,
+								declared_var->var_name.string);
+				print_parse_message(WARNING, info, parser_line_num);
+			}
+
+			//Fall through
+		case TYPE_CLASS_STRUCT:
 			//Actually create the node now
 			declaration_node = ast_node_alloc(AST_NODE_TYPE_DECL_STMT, SIDE_TYPE_LEFT);
 
--- a/oc/test_files/initialize_immut_pointer.ol
+++ b/oc/test_files/initialize_immut_pointer.ol
@@ -1,0 +1,17 @@
+/**
+* Author: Jack Robbins
+* Test the case where we want to initialize an immutable pointer
+*/
+
+pub fn main() -> i32{
+	//Immutable pointer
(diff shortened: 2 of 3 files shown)


jackr276:
Validated:
array_access_immut_pointer.ol
array_access_mut_pointer.ol
assign_to_immut_ptr.ol
assign_to_immutable_array.ol
casting_immutable_to_immutable.ol
casting_mutable_to_mutable.ol
initialize_immut_pointer.ol
invalid_immut_ptr_to_mut_ptr.ol
invalid_immut_reference_to_mut_ptr.ol
mutability_cast_violation.ol
mutable_struct_immutable_field.ol
mutate_after_init.ol
mutate_after_init_param.ol
mut_pointer_immut_array.ol
reassign_immut_pointer.ol
void_declared_variable.ol
void_in_array.ol
void_in_struct.ol
void_in_union.ol
void_mutable_return_type.ol

PR #436, opened 2025-11-27
[Front-End]: Taking the address of an array, stack or union should now function properly

### LINKED ISSUE #435: Taking the address of an array is not working
Currently in array_access_deref we do:

```
let arr:mut i32[] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];

	let x:mut i32** = &arr;

	//TODO THIS IS NOT WORKING
	@access_array(x);
```

It generates:
```
movq %rsp, %rdi
call access_array /* --> %eax */
```

It should really generate something like:
```
movq %rsp, 40(%rsp)
movq 40(%rsp), %rdi
call access_array
```
There should  be a step where we load the actual base address itself into memory

jackr276:
[Front-End]: Taking the address of an array, stack or union should now function properly

Closes #435

### DIFF
--- a/oc/compiler/stack_data_area/stack_data_area.c
+++ b/oc/compiler/stack_data_area/stack_data_area.c
@@ -6,7 +6,6 @@
 */
 
 #include "stack_data_area.h"
-#include <stdint.h>
 #include <stdio.h>
 #include <sys/types.h>
 #include <stdlib.h>


jackr276:
Validated:
array_access_deref.ol
array_access_mut_pointer.ol
array_address_double.ol
array_pointer_intermediary.ol
array_reference_double.ol
invalid_array_manipulation.ol
invalid_array_reassignment.ol
invalid_let_array_assignment.ol
struct_reference_double.ol

jackr276:
All looks good, we're ready to go.

PR #440, opened 2025-11-29
[Efficiency]: The lex_stack now uses an internal array, saving us on space and memory search complexity

### LINKED ISSUE #432: Lexstack should be using an array internally, not a linked list
Lexstack should be using an array internally, not a linked list

The linked list is unnecessarily heavy with pointers and allocations. We should just be using a simple dynamically resizing array.

jackr276:
[Efficiency]: The lexer stack now uses an internal array, saving us on space and memory search complexity

This change also removes all of the *_NOT_EMPTY and *_EMPTY constants in favor for the universal TRUE and FALSE constants.

Closes #432


jackr276:
Testing done, no changes to output and no memory errors. We are good to go.

PR #443, opened 2025-11-30
[Compiler Utilities]: Priority queue now has minimum and maximum versions, and test coverage

### LINKED ISSUE #442: Priority Min & Max queue implementations
Fix priority queue module to be a max-priority queue:

We currently use a min-heap. We need to change this to use a max-heap

jackr276:
[Compiler Utilities]: Priority queue now has minimum and maximum versions, and test coverage

Closes #442


jackr276:
All that we've had happen here is the tie-breaking scheme for the register allocator changed. This results in some differences in spilling and register allocation, but no fundamental differences.

PR #445, opened 2025-12-01
[Instruction Selector]: Rethinking of the Instruction DAG and rework of it's functionality with the new design

### LINKED ISSUE #444: Rework the Instruction DAG so that it fits the needs that we have
Rework the Instruction DAG so that it fits the needs that we have. Mainly:
fast topological sort
unidirectional
weighted

Currently it doesn't really fit any of these so we'll need to fix that.

### LINKED ISSUE #446: "Tie-Breaker" function needed for priorities with instructions
"Tie-Breaker" function needed for priorities with instructions:

We currently use the length of the longest path to a root for priority in the scheduler. This is good, but we do get a lot of ties so we need a "tie-breaker" function to decide who wins in the case of a match.

### LINKED ISSUE #447: Special Cycle Calculations for load instructions
Special Cycle Calculations for load instructions:

Since load instructions need to account for the possibility of a cache miss(in which case they will take 100s or 1000s of cycles to complete), we need to implement a special cycle calculation for them. The same issue does not exist for stores, because we are not reading from the cache there.

### LINKED ISSUE #448: Priority Queues need the option to tie break
Priority Queues need the option to tie break:

We should have a function pointer that allows us to tie break any two given nodes in the priority queue. The priority queue needs to optionally support this and we would need another test suite to cover it.

jackr276:
[Instruction Selector]: Rethinking of the Instruction DAG and rework of it's functionality with the new design

Closes #444 
Closes #446 
Closes #447
Closes #448

### DIFF
--- a/oc/compiler/data_dependency_graph/data_dependency_graph.c
+++ b/oc/compiler/data_dependency_graph/data_dependency_graph.c
@@ -562,6 +562,9 @@ static dynamic_array_t* get_all_connected_components(dynamic_array_t* subgraph, 
 
 		//Otherwise it hasn't been visited, so find it's connected components
 		connected_component_rec_DFS(node, connected_component);
+
+		//Once we're here, we can add this connected component to our overall array of them
+		dynamic_array_add(connected_components, connected_component);
 	}
 
 	//Give back the dynamic array of dynamic arrays of connected components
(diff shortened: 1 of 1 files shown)


jackr276:
This is a good initial first-cut of our entire module. There will need to be improvements/add ons as we go, but we've laid the foundation here for the scheduler.

PR #475, opened 2025-12-15
[Front End]: constant simplification subsystem has been added - now the parser will recognize and automatically handle cases where we can do constant arithmetic

### LINKED ISSUE #389: Constant simplification does not work properly
Look at an example like: lex_test11.ol

```
	*my_arr = 2+3 + 6-1;
	ret;
```

We end up not simplifying properly. Ideally this would even be handled by the parser.

### LINKED ISSUE #473: Move constant simplification, where possible, to the parser
Move constant simplification, where possible, to the parser: This is possible to do for all of our operations - adding, multiplying, dividing, etc. If the user does something like 5 + 1, or something like typesize(int) << 2, we should simplify that at the parser end.

### LINKED ISSUE #474: Constant simplification at the parser level needs test coverage
Constant simplification at the parser level needs test coverage: now that we have the ability to do this at the parser level, we need test coverage to accurately cover a wide enough range of use cases.

### LINKED ISSUE #476: Constant coercion in the parser
Constant coercion in the parser: The parser currently performs type coercion, but if it coerces the type of a constant, it will not coerce the constant itself. This is going to be important later for when we do floats, so we should just add this in now for non-floats while we're looking at the constant subsystem.

### LINKED ISSUE #478: Complex constant simplification test cases for order of operations
Complex constant simplification test cases for order of operations - we need test coverage to determine whether the order of operations is being well and truly followed with our parser level constant simplification

jackr276:
[Front End]: constant simplification subsystem has been added - now the parser will recognize and automatically handle cases where we can do constant arithmetic

Closes #389 
Closes #473 
Closes #474 
Closes #476 
Closes #478

### DIFF
--- a/oc/test_files/constant_simplification_shifting.ol
+++ b/oc/test_files/constant_simplification_shifting.ol
@@ -1,0 +1,28 @@
+/**
+* Author: Jack Robbins
+* Test the parser's ability to do in-flight constant calculations
+* with addition and subtraction
+*/
+
+define struct my_struct{
+	x:mut i32*;
+	y:i16;
+} as custom_struct;
+
+pub fn shift_type_size() -> i64{
+	ret typesize(custom_struct) << 2;
+}
+
+pub fn expanding_constant_size() -> i64{
+	ret 3 << 3;
+}
+
+pub fn shift_type_size2() -> i32 {
+	ret typesize(custom_struct) >> 2;
+}
+
+pub fn main() -> i32 {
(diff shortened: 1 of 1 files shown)


jackr276:
This is shaping up to be a 4000+ line MR. The reason for the size is that we have specialized processing methods for every single operation with constants. This leads to code duplication, but it is more comprehensible than having code in one gigantic function with 3 or 4 levels of nested switch statements and faster too, since we know what function to call

PR #490, opened 2025-12-22
[Intermediate Representation]: Completely remove the "memory address of <>" statements

### LINKED ISSUE #486: Rework/removal of "memory address of" statements
Rework/removal of "memory address of" statements:

After doing some reviews of the code I feel like this "memory address of" concept in the code is holding us back a lot. We don't need to add this in, we can just emit a variable that is a special MEM<<var-name>> variable as our representation and drop this nonsense with the memory address of altogether

### LINKED ISSUE #491: More complete use & integration of lea statements in OIR
More complete use & integration of lea statements in OIR:

These are currently completely underused, and really should be used more on the front end to make our address simplification easier on the backend

### LINKED ISSUE #492: Complete rewrite of the lea IR to use different addressing modes
Complete rewrite of the lea IR to use different addressing modes:

Currently the lea IR is very anemic, and doesn't support multiple addressing modes easily. We will fix this and this will also simplify our process of address simplification in the selector.

### LINKED ISSUE #493: Fix instruction simplifier rules for lea statements
Fix instruction simplifier rules for lea statements: now that we have a more expansive use of the lea statement, we can rewrite a bunch of simplifier rules to accomodate this fact and get less instructions overall

### LINKED ISSUE #494: Add support for index only scaled and index + displacement
Add support for index only scaled and index + displacement:

```
leaq (, %rcx, 2), %rax

// or

leaq 16(, %rcx, 4), %rax
```

This needs to be supported in *both* OIR form and in our generated assembly

### LINKED ISSUE #495: Lea scale factors may only go up to 8
Lea scale factors may only go up to 8:

I haven't seen this be an issue yet, but let's build in the protections now

### LINKED ISSUE #498: Additional simplification/compression rules for loads/stores that have lea's right beforehand
Additional simplification/compression rules for loads/stores that have lea's right beforehand. These additional simplification rules will allow us to have the same level of instruction compactness that we did before this big change

### LINKED ISSUE #499: Error codes for unselected instruction
Error codes for unselected instruction:

subq $32, %rsp
movb $97, (%rsp)
movl $2, 4(%rsp)
movl $3, 8(%rsp)
movl $4, 12(%rsp)
Not yet selected. Statement code is: 21
movl 4(%rsp, %rdi, 4), %eax
addq $32, %rsp
ret /* --> %eax */

In array_in_struct_initializer.ol. This is a bug and must be dealt with.

### LINKED ISSUE #500: Memory read/writes should not rely on base address types. They should instead be stored in the instruction itself
Memory read/writes should not rely on base address types. They should instead be stored in the instruction itself:

We should not be doing "type adjusted base addresses". Instead, store the memory read/write type in the store instruction itself and use that during instruction selection to avoid this confusing nonsense with base addresses.

jackr276:
[Intermediate Representatoin]: Completely remove the "memory address of <>" statements

Closes #486 
Closes #491 
Closes #492
Closes #493 
Closes #494 
Closes #495 
Closes #498 
Closes #499 
Closes #500

### DIFF
--- a/oc/test_files/address_of_global_variable.ol
+++ b/oc/test_files/address_of_global_variable.ol
@@ -19,13 +19,13 @@ pub fn main(argc:i32, argv:char**) -> i32 {
 	global_var = argc;
 
 	//Address of a global var
-	let x:mut i32* = &global_var;
-	
-	//Try to add to it
-	x = x + 3;
+	let x:mut i32* = &global_var + 3;
 
 	//Try to offset it
 	@ret_param(x[argc]);
+
+	//Test a bin op
+	@test_func(&global_var + argc);
 
 	ret @test_func(x);
 }


jackr276:
All looking good, ready to begin verifying.

jackr276:
Memory check clears:
image

jackr276:
Validated:
address_of_array_member.ol
address_of_global_variable.ol
array_testing.ol
duplicate_array_type_lookup.ol
pointer_math.ol
short_circuit.ol

jackr276:
We are looking good here. On the benefits side, this change to the IR has given us an IR that does what used to take 2 instructions:
```
t4 <- memory address of x
store t4[5] <- t5
```

And now does it in 1:
```
store MEM<t4> <- t5
```

This saves us on computation time for block traversal and overall memory usage

PR #504, opened 2025-12-28
[Optimizer]: Stack memory is now optimized using the same mark-and-sweep scheme as register variables

### LINKED ISSUE #381: Memory address/memory access optimization
Once the new memory model has been completed, it should simplify the way in which we handle memory access in some ways. Opening this for tracking to start the enhancement.

### LINKED ISSUE #485: Memory/stack region use count tracking
Memory/stack region use count tracking: with the new memory model that was added in October, we should be able to track uses for stack regions. This would enable us to automatically remove regions from the stack that were no longer used.

Another idea: if we have variables that are forced onto the stack(have their address taken), there may be a way to figure out if that is actually needed or not. If it's not, we could "de-stack" them and make them entirely register variables.

jackr276:
[Optimizer]: Stack memory is now optimized using the same mark-and-sweep scheme as register variables

Closes #381
Closes #485

### DIFF
--- a/oc/compiler/optimizer/optimizer.c
+++ b/oc/compiler/optimizer/optimizer.c
@@ -298,9 +298,6 @@ static void mark_and_add_definition(cfg_t* cfg, three_addr_var_t* variable, symt
  * 				add j to worklist
  */
 static void mark(cfg_t* cfg){
-	//Pointer to the current function record
-	symtab_function_record_t* current_function;
-
 	//First we'll need a worklist
 	dynamic_array_t worklist = dynamic_array_alloc();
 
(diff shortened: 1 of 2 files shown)


jackr276:
Ready to begin validations. Ollie memory check clears:
image

jackr276:
Validated:
array_access_deref.ol
array_testing.ol
constant_simplification_add_subtract.ol
constant_simplification_bitwise_and.ol
constant_simplification_equals.ol
constant_simplification_exclusive_or.ol
constant_simplification_inclusive_or.ol
constant_simplification_not_equals.ol
constant_simplification_relational_operators.ol
constant_simplification_shifting.ol
construct_pointer_access.ol
duplicate_array_type_lookup.ol
lex_test10.ol
multiplicative_constant_expressions.ol

jackr276:
Validations so far looking good

PR #509, opened 2025-12-29
[Compiler Utilities][Performance]: Symtab hashing scheme has been written to optimize for size & speed

jackr276:
[Compiler Utilities][Performance]: Symtab hashing scheme has been written to optimize for size & speed

We have abandoned the old method and implemented FNV-1a hashing, with smaller keyspaces


jackr276:
No test files changed, only the order of variable/function warning printing

PR #510, opened 2025-12-30
[Intermediate Representation]: Rewrite of Local Constants to support RIP relative addressing, more than just strings

### LINKED ISSUE #454: Returning a string literal needs to be tested and work
Returning a string literal needs to be testing and work:

We need to be able to return a string literal upon a char* return type. This should work because string literals are simply local constants inside of the assembly file.

### LINKED ISSUE #508: Complete rewrite of local constants
Currently local constants only support strings and do not allow us to take advantage of rip-relative addressing modes at all. This should be entirely rewritten so that local constants can take any kind of constant, and so that they can be used as variables(like global vars) in rip-relative addressing.

jackr276:
[Intermediate Representation]: Rewrite of Local Constants to support RIP relative addressing, more than just strings

Closes #454 
Closes #508

### DIFF
--- a/oc/compiler/file_builder/file_builder.c
+++ b/oc/compiler/file_builder/file_builder.c
@@ -21,6 +21,9 @@ static void print_assembly_block(FILE* fl, basic_block_t* block){
 		//First we print out the local constants that the block has
 		print_local_constants(fl, block->function_defined_in);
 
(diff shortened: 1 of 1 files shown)


jackr276:
TODO: returning a string literal needs to be tested

jackr276:
All looks good after review, we are clear for merge

PR #513, opened 2025-12-31
[Front-End]: Add support for f32 and f64 constant values inside of the ollie compiler

### LINKED ISSUE #512: Floating point constants support
Floating point constants support: floating point constants will be represented in local constants, not by immediate values, so this change will be something other than just adding the area into the constant module.

jackr276:
[Front-End][Back-End]: Add support for f32 and f64 constant values inside of the ollie compiler

Closes #512


jackr276:

image

PR #516, opened 2026-01-01
[Front End][Back End]: f32, f64 constants are now fully supported, constant system now supports: f32, f64, i16, u16

### LINKED ISSUE #186: Floating point support
Complete support for floating point values & operations in ollie.

This will likely be an enormous change, and as such this will not be done until after everything else in the backlog is complete.

### LINKED ISSUE #515: Floating point global variable initialization support
Floating point global variable initialization support: we need to support users initializing/using f32/f64 global variables. This is mainly a front-end thing but it would extend all throughout the compiler. They will likely be handled similarly to the way that floating point constants themselves are handled

### LINKED ISSUE #519: Short constant specifier(s suffix) is needed in the lexer
Short constant specifier(s suffix) is needed in the lexer: support and a few test cases are needed for this

jackr276:
[Front End][Back End]: f32,f64,i16,u16 constants are now fully supported throughout the system

Closes #186 
Closes #515 
Closes #519

### DIFF
--- a/oc/compiler/ast/ast.c
+++ b/oc/compiler/ast/ast.c
@@ -5162,17 +5162,29 @@ void less_than_or_equal_to_constant_nodes(generic_ast_node_t* constant_node1, ge
 		case INT_CONST_FORCE_U:
 			//Now go based on the second one's type
 			switch(constant_node2->constant_type){
+				case DOUBLE_CONST:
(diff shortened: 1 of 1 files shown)


jackr276:
All looks good for this MR, ready to merge

PR #520, opened 2026-01-04
[Type System]: F32 and F64 fully integrated into the type assignability checker, coercer

### LINKED ISSUE #450: Review of pointer assignability rules with integers
Review of pointer assignability rules with integers:

Currently we force a u64. I am unsure if that's the right way to go always or we should keep it. More thought towards this/investigation is needed.

### LINKED ISSUE #471: Review of handling/rewrite of pointers to arrays & assignability
Review of handling/rewrite of pointers to arrays & assignability: currently we allow things like "i32** = i16[]*". This should definitely fail. We also need to think about how this really works and fix/add test coverage where needed. Booking this top level issue to do so.

### LINKED ISSUE #514: Floating point type system support
Floating point type system support: we need to fully support floating point operations throughout the type system. This includes constant coercion, type assignability and compatability with operations.

### LINKED ISSUE #521: Representation of array decay into equivalent pointers is required
Representation of array decay into equivalent pointers is required: From what I can tell, this is either not implemented in the type system or it's implemented incorrectly. This needs to be looked at.

jackr276:
[Type System]: F32 and F64 fully integrated into the type assignability checker, coercer

Closes #450 
Closes #471
Closes #514
Closes #521

### DIFF
--- a/oc/compiler/type_system/type_system.c
+++ b/oc/compiler/type_system/type_system.c
@@ -1591,7 +1591,6 @@ u_int8_t is_binary_operation_valid_for_type(generic_type_t* type, ollie_token_t 
 		case G_THAN_OR_EQ:
 		case NOT_EQUALS:
 		case DOUBLE_EQUALS:
-		case PLUS:
 			switch(type->type_class){
 				//Basic types(minus void) work
 				case TYPE_CLASS_BASIC:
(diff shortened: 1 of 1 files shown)


jackr276:
Validated:
array_address_arithmetic.ol
array_arithmetic.ol
assign_float_to_int.ol
coerce_constant_float_int.ol
integer_to_pointer.ol
invalid_larger_pointer_to_smaller_array.ol

PR #524, opened 2026-01-05
[Back End Codegen]: Floating point registers added in, beginning of an internal split between general purpose and SSE processes

### LINKED ISSUE #511: Floating point register support & basic functionality
Floating point register support & basic functionality - we need backend support for the 16 xmm registers for floating point operations.

### LINKED ISSUE #523: Support in three_addr_var_t for both SSE and gen purpose registers
Support in three_addr_var_t for both SSE and gen purpose registers: we need the ability to store these both, probably in a union since they are mutually exclusive.

### LINKED ISSUE #525: Completely remove all "block_terminal_type" logic in the CFG
This is a remnant of the old way we did things in the CFG and serves no real purpose for us now. We should remove this entirely for maintainability and code style reasons.

jackr276:
[Back End Codegen]: Floating point registers added in, beginning of an internal split between general purpose and SSE processes

Closes #511 
Closes #523 
Closes #525

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -87,7 +87,7 @@ typedef struct{
  * Determine whether a given block needs to have a jump appended to it
  * If the block is empty *or* it doesn't end in a return, we need a jump
  */
-#define IS_ENDING_JUMP_REQUIRED(block) (((block->exit_statement == NULL)\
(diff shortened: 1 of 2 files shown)


jackr276:
We will also use this to remove the block_terminal_type_t feature in the CFG since this MR is already open and it's a small internal fix

PR #533, opened 2026-01-08
[Register Allocator]: Floating point registers are now fully supported in the allocator

### LINKED ISSUE #529: Complete rewrite of the register allocator
Complete rewrite of the register allocator: the register allocator needs to be biforcated into floating point & non-floating point segments. This allows the floating point and non-floating point areas of the code to have separate interference graphs, spill crawls, etc.

### LINKED ISSUE #531: Precoloring for floating point registers
Precoloring for floating point registers: we need to handle precoloring for this unique subset of registers. It should be done in the same crawl as the regular instruction precoloring for efficiency.

### LINKED ISSUE #532: Separate spill processes for float, non-float registers
Separate spill processes for float, non-float registers: we need to have separate spill processes for float and non-float registers. This means that if a non-float register is spilled, the *only* graph that we should be redoing is the float graph, not the entire graph.

### LINKED ISSUE #534: Separate live now & live after functions are needed for SSE vs non-SSE live ranges
Separate live now & live after functions are needed for SSE vs non-SSE live ranges. We will need to compute these values initially *and* anytime we have a coalescence or spill. To make things more efficient, we should only be calculating live after/live now if we have to for each class of register.

### LINKED ISSUE #540: Caller/callee saving for floating point registers
Caller/callee saving for floating point registers: floating point registers need to have caller saving fully implemented. In Ollie, functions may clobber the XMM registers however they please, so everything is a caller save.

### LINKED ISSUE #541: Custom allocator function just for floating point registers
Custom allocator function just for floating point registers: we need a custom allocator function for exclusively floating point registers. This will need to link into the spiller just like the one for general purpose registers does

### LINKED ISSUE #542: Bitmap instead of boolean array for register allocation
Currently, we use a boolean array that needs to be allocated on the stack and memset to 0 every time. We should instead be using a 32 bit integer, set to 0, and then indexing into it using bitwise logic. This makes the entire process register-local and does not require us to use any stack memory in what is a hot-zone in the register allcoator.

jackr276:
[Register Allocator]: Floating point registers are now fully supported in the allocator

Closes #529
Closes #531
Closes #532 
Closes #534 
Closes #540 
Closes #541 
Closes #542

### DIFF
--- a/oc/test_files/float_caller_saving.ol
+++ b/oc/test_files/float_caller_saving.ol
@@ -1,0 +1,34 @@
+/**
+* Author: Jack Robbins
+* Test a more involved f32 add/subtract program for register 
+* allocation
+*/
+
+fn manipulate_float_values(x:f32, y:f32) -> f32 {
+	let result:mut f32 = x + y;
+
+	let z:mut f32 = x - y;
+
+	let result_helper:f32 = x - (y + z);
+
+	result -= z;
+	
+	ret result + result_helper;
+}
+
+
+fn float_add_tester(x:f32, y:f32, z:f32) -> f32 {
+	let result:mut f32 = x + y;
+
+	let result_helper:f32 = x - (y + z);
+
+	result -= @manipulate_float_values(y, z);
+	
+	ret result + result_helper;
+}
+
+
+
+pub fn main() -> i32 {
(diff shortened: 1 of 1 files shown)


jackr276:
We appear to have restored full functionality of the non-sse module

PR #536, opened 2026-01-10
[Front-End]: boolean true & false constants, u8 & i8 constant type support

### LINKED ISSUE #517: Only private functions should generate warnings about not being called
Only private functions should generate warnings about not being called: currently all functions do this and it's not accurate, because public functions are exposed via the linker and can be called by other files

### LINKED ISSUE #518: Constant system should have signed, unsigned i8/u8 values for byte values as opposed to the current "char_value"
Constant system should have signed, unsigned i8/u8 values for byte values as opposed to the current "char_value". This will allow for greater flexibility. In Ollie, a char is a signed 8-bit integer for reference.

### LINKED ISSUE #535: True & False constants in the lexer/parser
We need support for "true" and "false" constants in the front-end. This is just a nice to have and we should do this since we already support booleans.

### LINKED ISSUE #537: Test coverage for new true/false constants
Test coverage for new true/false constants: we should have a small bit of test coverage for this, nothing crazy but we need functionality verification

jackr276:
[Front-End]: boolean true & false constants, u8 & i8 constant type support

Closes #517 
Closes #518 
Closes #535 
Closes #537

### DIFF
--- a/oc/compiler/lexer/lexer.c
+++ b/oc/compiler/lexer/lexer.c
@@ -22,7 +22,7 @@
 #include "../utils/constants.h"
 
 //Total number of keywords
-#define KEYWORD_COUNT 50
+#define KEYWORD_COUNT 52
 
 //We will use this to keep track of what the current lexer state is
 typedef enum {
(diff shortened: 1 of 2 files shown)


jackr276:
Validated:
true_false_constants.ol

jackr276:
This MR is looking good. Extensive testing is not done because of how long/large it would be, but this will suffice for now

PR #544, opened 2026-01-13
[Lexer]: remove/rework to avoid the need for the "SEARCHING_FOR_CONSTANT" flag, fixes to Ollie enum parsing

### LINKED ISSUE #543: Completely deprecate the "searching for constant" flag in the lexer
Completely deprecate the "searching for constant" flag in the lexer: This, from what I can tell, is just a mistaken implementation and should not be used at all. We should instead tokenize as follows:

x - -9 as -> "x" "MINUS" "MINUS" "9" - this would become something like:
```     
    - 
x    -  
        9
```

We should be negating constants in the parser, not in the lexer. Once this is done, our design flaw will be solved

### LINKED ISSUE #545: Rework of how positive/negative constants are handled
Rework of how positive/negative constants are handled: we now need to rework how we handle seeing "+/-" tokens in front of constants where appropriate. This used to be handled by the lexer with the "SEARCHING_FOR_CONSTANT" flag, but this will now need to be a parser side evaluation.

### LINKED ISSUE #546: Complete rework of case statement handling
Complete rework of case statement handling: we need a complete rework of case statement handling, especially enum handling, and how we handle generating warnings and such.

Testing is also needed for the user to do something like:

case typesize(my_struct) for example. It should work

### LINKED ISSUE #547: Type assignability with enums
Type assignability with enums: we should generate errors for enums if:

The user tries to assign a number that is not in the enum range
The user tries to assign one enum value to another

### LINKED ISSUE #548: Complete overhaul of enum types in the parser
Complete overhaul of enum types in the parser: I'm not sure how this slipped past, but it looks like enum types are entirely cobbled together. What really should happen at a base level is any time an enum member identifier is come across, it should be automatically converted into a constant by the parser and treated as such. Enums are at the end of the day just fancier ways of writing constants.

jackr276:
[Lexer]: remove/rework to avoid the need for the "SEARCHING_FOR_CONSTANT" flag

Closes #543 
Closes #545 
Closes #546 
Closes #547 
Closes #548

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -1129,27 +1129,6 @@ static generic_ast_node_t* function_call(FILE* fl, side_type_t side){
 				return print_and_return_error(info, parser_line_num);
 			}
 
-			//If this is a constant node, we'll force it to be whatever we expect from the type assignability
-			if(current_param->ast_node_type == AST_NODE_TYPE_CONSTANT){
-				current_param->inferred_type = final_type;
-
-				//Do coercion
-				perform_constant_assignment_coercion(current_param, final_type);
-			}
-
-			//Special checking here - if we have an enum type that is being assigned to, we need
-			//to make sure that it's being assigned to a valid value in it's range
-			if(is_enum_type(param_type) == TRUE && current_param->ast_node_type == AST_NODE_TYPE_CONSTANT){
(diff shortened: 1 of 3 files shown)


jackr276:
It looks to me that this effectively "solves" our enum problems, at least in the current scope of the language. Enums have been an afterthought for a while, so with this change we're bringing them in as first-class citizens. This is ready to be validated and merged

PR #550, opened 2026-01-15
[Front End/Instruction Selector]: Special logic for floating point increment and decrement operations added 

### LINKED ISSUE #527: Floating point increment & decrement
Floating point increment & decrement: unlike with basic instructions, there is no "inc" or "dec" for floating point. It will need to be converted into a +/- one in a similar way to how pointers are currently handled. This would need to happen on the front end.

### LINKED ISSUE #551: [Triage]: floating point constants are completely incorrect
[Triage]: floating point constants are completely incorrect: it has come to my attention that what we are doing for floating point constants(just using lea) is completely incorrect and will need to be updated

### LINKED ISSUE #552: Combining load/store with rip-relative addressing in the simplifier
Combining load/store with rip-relative addressing in the simplifier: we need to be using the simplifier to combine something like:

```
t4 <- lea .LC1(rip)
t5 <- (t4)
```

into 

```
t5 <- .LC1(rip)
```

jackr276:
[Front End]: Special logic for floating point increment and decrement operations added

Closes #527 
Closes #551 
Closes #552

### DIFF
--- a/oc/test_files/float_add_subtract_involved.ol
+++ b/oc/test_files/float_add_subtract_involved.ol
@@ -4,6 +4,8 @@
 * allocation
 */
 
+SABOTAGED
+
 fn float_add_tester(x:f32, y:f32, z:f32) -> f32 {
 	let result:mut f32 = x + y;
 
--- a/oc/test_files/float_caller_saving.ol
+++ b/oc/test_files/float_caller_saving.ol
@@ -4,6 +4,8 @@
 * allocation
 */
 
+SABOTAGED
+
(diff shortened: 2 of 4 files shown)


jackr276:
Note that there are several test cases that have had "SABOTAGED" appended to the front to stop them from compiling/causing errors. This will be removed *once* we have the floating point load/store implemented. Until then, we will be leaving it like this.

jackr276:
Validated:
postoperations_on_constants.ol
preinc_dec_constants.ol

PR #556, opened 2026-01-18
[Triaged Fix]: float caller saving now correctly uses the stack instead of push/pop instructions

### LINKED ISSUE #555: [Triage] Float caller saving is generating incorrect assembly
[Triage] Float caller saving is generating incorrect assembly: We are not able to push/pop the floating point registers. We must either push/pop quad-word registers as a proxy for them, or we could just save them to the stack when need be.

jackr276:
[Triaged Fix]: float caller saving now correctly uses the stack instead of push/pop instructions

Closes #555


jackr276:
Memory check coming back clean:

image

jackr276:
Validated:
float_caller_saving.ol
float_caller_saving_indirect.ol

PR #586, opened 2026-01-28
[Instruction Selector]: lea statement simplifier has been enhanced to recognize & simplify 1, 0 multipliers, load simplifier logic has been enhanced

### LINKED ISSUE #584: Simplify addressing mode operations where the lea multiplier is 1
Simplify addressing mode operations where the lea multiplier is 1: there is at least one example of this that I've seen:

```
movw $3, (%rcx, %rdx, 1)
```

There's no need to keep the 1 around so we should simplify it away. This will probably also lead to more opportunities for simplification

### LINKED ISSUE #587: Redundant copy folding for addressing mode instructions
Redundant copy folding for addressing mode instructions: we need to leverage the instruction selector to do redundant copy folding for some addressing mode expressions. Here's an example:

```
movq ^t2_0, t3
movb (^t1_0, t3), t4
```

Though in theory this should be picked up by the coalescer, in reality it will not be because many times the source variable(in this case ^t2_0) is "live_out" of the block and therefore technically interferes with t3 and cannot be coalesced. This is not an issue with the coalescer, and it is in fact working as intended here. We will need to do this ourselves in the selector.

### LINKED ISSUE #588: Enhanced simplifier logic for loads
Enhanced simplifier logic for loads: currently we have a whole array of simplifier logic target around stores, including copy folding and binary operation folding. We need to be doing the exact same for load operations

### LINKED ISSUE #589: Enhanced copy folding simplifier logic for stores
Enhanced copy folding simplifier logic for stores: we just did this for load instructions with great results, reducing instructions & register interference. We need to do the exact same for store instructions as it is not currently in.

jackr276:
[Instruction Selector]: lea statement simplifier has been enhanced to recognize & simplify 1, 0 multipliers, load simplifier logic has been enhanced

Closes #584 
Closes #587 
Closes #588
Closes #589

### DIFF
--- a/oc/compiler/instruction_selector/instruction_selector.c
+++ b/oc/compiler/instruction_selector/instruction_selector.c
@@ -3063,12 +3063,28 @@ static u_int8_t simplify_window(instruction_window_t* window){
 
 				break;
 
+			//Opportunity to copy fold here
 			case THREE_ADDR_CODE_ASSN_STMT:
 				//These specific conditions must be met for this to be true
 				if(preceeding_instruction->assignee->variable_type == VARIABLE_TYPE_TEMP
 					&& preceeding_instruction->assignee->use_count == 1
 					&& variables_equal(preceeding_instruction->assignee, store_instruction->op1, FALSE) == TRUE){
 				}
+
+				//Unlike above, we do not need to change the statement type. We just need to shuffle around the values
+				//that we already have
+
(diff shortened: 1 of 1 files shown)


jackr276:
Validated:
combined_jump_endings.ol
duplicate_string_local_constants.ol
lea_simplification_with_1_multiplier.ol

PR #594, opened 2026-01-31
[Front-End]: function calling logic has been revised to be more efficient and complete, individual call graph module is deprecated, inline keyword added

### LINKED ISSUE #591: Recursive function detection
Recursive function detection: we need the ability to detect whether or not a function is calling itself. This is going to be important for when we eventually do function inlining, as we are going to stop recursive functions from being inlined.

### LINKED ISSUE #593: Completely deprecate the call-graph, functionality needs to be moved entirely to the symtab
Completely deprecate the call-graph, functionality needs to be moved entirely to the symtab

### LINKED ISSUE #595: New utility class "dynamic_set"
New utility class "dynamic_set" - this allows for uniqueness enforcement on pointers in a dynamic array. It should be a separate module because we don't want to start bolting on extra functionality to the regular darray

### LINKED ISSUE #596: "inline" keyword addition and parser-side support
"inline" keyword addition and parser-side support: we want this inside of the parser only for now, along with any needed restrictions. CFG support will be a large effort and will come later

jackr276:
[Front-End]: function calling logic has been revised to be more efficient and complete

Closes #591 
Closes #593 
Closes #595 
Closes #596

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -5099,9 +5099,9 @@ static u_int8_t function_pointer_definer(ollie_token_stream_t* token_stream){
 	push_token(&grouping_stack, lookahead);
 
 	//Once we've gotten past this point, we're safe to allocate this type. Function
-	//pointers are always private
-	generic_type_t* mutable_function_type = create_function_pointer_type(FALSE, parser_line_num, MUTABLE);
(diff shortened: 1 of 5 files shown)


jackr276:
Validated:
basic_inline_function.ol
fibonacci_recursive.ol
indirect_function_calls_warnings.ol
indirect_recursion.ol
invalid_direct_recursive_inline.ol
invalid_indirect_recursion_inline.ol
invalid_inline_forward_declare_mismatch.ol
invalid_public_inline.ol

PR #615, opened 2026-02-08
[Instruction Selector]: handling has been added to support moving bytes/shorts into float/double variables

### LINKED ISSUE #564: Handling for converting bytes/shorts to floating point
Handling for converting bytes/shorts to floating point: this is technically acceptable via the language spec. Odds are we will need to first convert such values into an i32/u32, and then use that as the float value.

### LINKED ISSUE #613: Byte/short to float - from memory conversion
Byte/short to float - from memory conversion: we need to test how this is handled when we're going from a byte/short in memory to a float

### LINKED ISSUE #614: Byte/short to float - to memory conversion
Byte/short to float - to memory conversion: we need a conversion step when we're moving items into memory(store)

jackr276:
[Instruction Selector]: handling has been added to support moving bytes/shorts into float/double variables

Closes #564 
Closes #613 
Closes #614


jackr276:
image

PR #618, opened 2026-02-08
[Register Allocator]: PXOR clear instructions no longer cause excessive interference

### LINKED ISSUE #616: Special live range handling for PXOR instructions with identical source/dest
Special live range handling for PXOR instructions with identical source/dest:

Currently, these instructions are causing the register allocator(specifically the liveness analyzer) to believe that the source was alive before the instruction. For instance:

```
pxor LR9, LR9
```

This is a special case for our allocation. LR9 actually was not alive beforehand, we are just zeroing it out. We need to figure out a way either in the instruction itself or at the live range level to recognize this and avoid seeing interference where there is none.

jackr276:
[Register Allocator]: PXOR clear instructions no longer cause excessive interference

Closes #616

### DIFF
--- a/oc/test_files/pxor_interference_test.ol
+++ b/oc/test_files/pxor_interference_test.ol
@@ -3,25 +3,18 @@
 * Test the PXOR interference in the register allocator
 */
 
-
-//Completely meaningless function
-pub fn do_something_float(x:f32) -> f32 {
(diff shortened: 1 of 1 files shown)


jackr276:
Validated:
pxor_interference_test.ol
store_floats.ol

PR #620, opened 2026-02-09
[Optimizer]: The optimizer will now recognize and simplify always true/false branches

### LINKED ISSUE #600: Simplification/optimization for always true/false scenarios
Simplification/optimization for always true/false scenarios:

The compiler needs to recognize and simplify things like:

```
if (x == x)
```

or 

```
if(1)
```

This would probably lead to a rewrite of the control flow itself

### LINKED ISSUE #621: Per-function operation in the optimizer
Per-function operation in the optimizer: the optimizer currently runs on a global basis. This is inefficient, especially when we have "while changed" algorithms where one change anywhere will trigger a rerun. We need to rewrite this to be on a per-function basis.

### LINKED ISSUE #622: Reworking of the user-defined unconditional jump
Reworking of the user-defined unconditional jump: currently this for some reason transforms itself into an always true conditional jump. This is nonsensical. We should rework how the parser/cfg does this.

### LINKED ISSUE #623: Always true/false optimization for CMP
Always true/false optimization for CMP: these are a bit more complex but we can still do it

### LINKED ISSUE #624: Conditional optimizations for logical not
Conditional optimizations for logical not when a value is always true/false

### LINKED ISSUE #625: Update to postprocessor to support branch removal if both targets are the same
Update to postprocessor to support branch removal if both targets are the same: we need to optimize scenarios like below:

```
cmpl $3,%edi
je .L10
jmp .L10

.L10:
cmpl $2,%esi
```

To just be jumps.

### LINKED ISSUE #628: Maintain an internal array of basic blocks for each function
Maintain an internal array of basic blocks for each function: we should just store this instead of having to rebuild this on the fly in two separate places(CFG & optimizer)

### LINKED ISSUE #629: CFG liveness, dominance need to be entirely per function
CFG liveness, dominance need to be entirely per function: this should give us some memory efficiency/processing boosts because currently, a lot of the dominance calculations are seeded with the entire CFG's blocks. This should all be done per-function

jackr276:
[Optimizer]: The optimizer will now recognize and simplify always true/false branches

Closes #600 
Closes #621 
Closes #622 
Closes #623 
Closes #624 
Closes #625 
Closes #628 
Closes #629

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -192,16 +192,18 @@ static inline void perform_constant_assignment_coercion(generic_ast_node_t* cons
 
 /**
  * Determine whether or not a variable is able to be assigned to
+ *
+ * A variable can only be assigned to if it is mutable. If the user
+ * does something like:
+ *
+ * declare x:i32;
+ * x = 5;
+ *
+ * This is inavlid, because you're violating the basic mutability constraint
  */
 static inline u_int8_t can_variable_be_assigned_to(symtab_variable_record_t* variable){
 	//Extract the type - it contains the mutability information
 	generic_type_t* type = variable->type_defined_as;
-
-	//If this hasn't been initialized then yes
-	//we can assign to it
-	if(variable->initialized == FALSE){
-		return TRUE;
-	}
 
 	//Otherwise, let's see if the type allows us to be mutated 
 	//If so - then we're fine. If not, then we fail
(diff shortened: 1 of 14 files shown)


jackr276:
Validated:
branch_reduction_if_always_false.ol
branch_reduction_if_always_true.ol
cmp_always_false_optimization.ol
cmp_always_true_optimization.ol
combined_jump_endings.ol
conditional_jump.ol
**confusing_jumps.ol**
constant_conditional_check.ol

VALIDATIONS FAILED -- we have more work to do, see comments in latest commit

PR #626, opened 2026-02-11
[Front End]: Update mutability rules for declarations

### LINKED ISSUE #598: Warnings when a user declares an immutable primitive
Warnings when a user declares an immutable primitive:

When a user does something like:

```
declare x:i32;
```

We should give a warning

### LINKED ISSUE #627: Update mutability logic for declared variables
Update mutability logic for declared variables: we should not be trying to guess whether a variable has been initialized or not. If the user does something like:

```
declare x:i32;

x = 5;
```

This should fail because it is not mutable. If it is not mutable, then you cannot assign to it, end of story.

jackr276:
[Front End]: Update mutability rules for declarations, reduce excessive warning messages

Closes #598 
Closes #627

### DIFF
--- a/oc/test_files/may_be_used_uninitialized.ol
+++ b/oc/test_files/may_be_used_uninitialized.ol
@@ -1,0 +1,21 @@
+/**
+* Author: Jack Robbins
+* 
+* Test the ability to detect if a variable may be used uninitialized
+*/
+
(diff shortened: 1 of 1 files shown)


jackr276:
This is looking good. We are set to merge here

PR #632, opened 2026-02-13
[Front-End]: the local constant system is no longer per function

### LINKED ISSUE #630: Local constants must not be stored on a per-function basis
Local constants must not be stored on a per-function basis: we should not be doing this for several reasons. The first being that it is inefficient. Second, it stops our optimizer from finding duplicate values across functions. And third, it is blocking our development of global char* variables that need the .LC system.

### LINKED ISSUE #631: Remove local constants from the symtab entirely
Remove local constants from the symtab entirely: it was likely a mistake to have them there in the first place. There should be an entirely separate local constant submodule that we are able to include/link to as needed

jackr276:
[Front-End]: the local constant system is no longer per function

Closes #630 
Closes #631


jackr276:
All results look good, we are set

PR #636, opened 2026-02-15
[Front-End]: Global variable char* and char[] variables are now supported

### LINKED ISSUE #574: Global variable char* and char[] variables
Global variable char* and char[] variables: we need to support these but the generated assembly is different than other global variables. Specifically a char*. We will need to investigate and implement

### LINKED ISSUE #634: Global char* initialization
Global char* initialization: this is also a special case. We will need to figure out how to support this as well.

### LINKED ISSUE #635: Ollie recursive macro subsistution
Ollie recursive macro subsistution: we need to ensure that recursive macros(macros with macro parameters) are supported in the preprocessor

jackr276:
[Front-End]: Global variable char* and char[] variables are now supported

Closes #574 
Closes #634
Closes #635

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -8851,6 +8851,33 @@ static void emit_global_array_initializer(generic_ast_node_t* array_initializer,
 
 
 /**
+ * This helper function is used to determine if we need to place a global variable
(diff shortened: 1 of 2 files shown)


jackr276:
image

jackr276:
This is looking good. There is a known bug with char*[] access but this is something that is going to be dealt with separately

PR #640, opened 2026-02-15
[Triaged Fix]: Pointer array access now works with proper intermediate memory loads

### LINKED ISSUE #637: [Triage]: accessing into arrays of pointers is incorrect
[Triage]: accessing into arrays of pointers is incorrect: we need to be emitting separate derefencing of the pointers when we do this. Currently this does not happen and it leads to generated code that is entirely incorrect

### LINKED ISSUE #638: Redefine how [] access works
Redefine how [] access works: bracket access is just syntactic sugar for *(ptr + offset). I think we've gotten too far away from what this really is and it's causing issues for us. We should try to refocus onto that itself.

### LINKED ISSUE #639: Concept of contiguous and non-contiguous memory region types
Concept of contiguous and non-contiguous memory region types: this needs to be added into the type system when we have something like int**, int*[], etc. These types are not contiguous("flat") regions of memory and if we are trying to get at them using the array accessor, then we'll need to do intermediary loads to do so.

jackr276:
[Triaged Fix]: Pointer array access now works with proper intermediate memory loads

Closes #637 
Closes #638 
Closes #639

### DIFF
--- a/oc/test_files/different_struct_access.ol
+++ b/oc/test_files/different_struct_access.ol
@@ -5,7 +5,7 @@
 
 define struct my_struct {
 	c:mut char;
-	x:i32[6];
+	x:mut i32[6];
 	y:i64;
 } as custom_struct;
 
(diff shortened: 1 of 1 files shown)


jackr276:
Memory check comes up clean:
image

jackr276:
We are ready to begin validations

jackr276:
It looks like we have the exact same issue here with struct access itself

jackr276:
Memory check looks good:
image

jackr276:
This issue has been resolved. The only existing issues that remain are how we handle array decay into pointers. With this new system, our old way will not work. However, that is out of scope for this and will be handled down the road

PR #650, opened 2026-02-19
[Type System]: Completely deprecate reference types

### LINKED ISSUE #467: References part 2 - additional refinement, new test files
Part 2 for references. Now that the baseline implementation is done, we need more testing and more use cases. These include
things like:
arrays of references
references in structs - do we support it or not
banning references in unions
references to pointers and pointers to references

All of these need to be considered, tested and dealt with accordingly. Booking this top level issue for tracking.

### LINKED ISSUE #649: Completely deprecate references in the type system
Completely deprecate references in the type system: References have proven to be a large headache for us. Since I never plan on using them, I think it's best if they get removed from the language entirely to reduce complexity and make us more focused on what really matters.

jackr276:
[Type System]: Completely deprecate reference types. The more I've used reference types, the less I like them. Pointers do the same thing with 1/4 of the complexity. We are completely deprecating them in Ollie

Closes #467 
Closes #649


jackr276:
Memory check comes up clean:
image

PR #654, opened 2026-02-21
[Type System]: overhaul of pointer and array assignability

### LINKED ISSUE #651: Type system support for pointers to arrays
Type system support for pointers to arrays: we need type-system level support for pointers to arrays like this:

```
let x:i32[5]*
```

This is in contrast to an array of pointers like:

```
let x:i32*[5]
```

This is currently not supported in the parser and will require some more work to get running

### LINKED ISSUE #652: Triage: parameter duplicate variable checking is broken
Triage: parameter duplicate variable checking is broken: currently this duplicate name checking does not work because the variable scopes are wrong. The function parameter list should be contained within the same variable scope as the outermost compound statement

### LINKED ISSUE #653: Array types need to be supported inside of the types_assignable function
Array types need to be supported inside of the types_assignable function: currently array types are only supported for char* to char[]. We should generalize this to all arrays. The gatekeeping for array types not being assignable will remain at the parser-level.

### LINKED ISSUE #658: Review of []* indexing inside of structs
Review of []* indexing inside of structs: booking this issue to have a generalized review/few test cases of this exact scenario.

### LINKED ISSUE #659: Indexing into array pointers
Indexing into array pointers: currently this seems incorrect. We'll need to do a thorough review and make it correct

### LINKED ISSUE #660: Triage: storing memory addresses does not work
Triage: storing memory addresses does not work:

Here's the issue:
```
store MEM<array_of_pointers_0>[0] <- MEM<arr1_0>
```

This will cause a live range error because we can't handle that RHS.

This should instead be
```
t4 <- MEM<arr1_0>
store MEM<array_of_pointers_0>[0] <- t4
```

jackr276:
[Type System]: overhaul of pointer and array assignability

Closes #651
Closes #652 
Closes #653 
Closes #658 
Closes #659 
Closes #660

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -4603,8 +4603,6 @@ static cfg_result_package_t emit_unary_operation(basic_block_t* basic_block, gen
 			//And the final type comes from when we dereference it
 			generic_type_t* dereferenced_type = dereference_type(pointer_type);
 
-			printf("DEREFED TYPE IS %s\n", dereferenced_type->type_name.string);
-
 			/**
 			 * If we what we have is an array pointer, then we don't need to do anything besides assign the
 			 * value over. This is because when we take the address of an array, all that we do is load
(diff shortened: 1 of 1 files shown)


jackr276:
image

PR #657, opened 2026-02-21
[Front End]: Floating point constants with a value of 0 are now initialized using PXOR

### LINKED ISSUE #655: Float 0.0 constants should be initialized using pxor
Float 0.0 constants should be initialized using pxor: we should not be using the constant subsystem when we can initialize these using pxor on the variable.

jackr276:
[Front End]: Floating point constants with a value of 0 are now initialized using PXOR

Closes #655


jackr276:

image

PR #678, opened 2026-03-01
[Optimizer]: Branch threading/jump hoisting implementation

### LINKED ISSUE #674: [Optimizer]: implement the branch copying step in the branch_reduce() algorithm
[Optimizer]: implement the branch copying step in the branch_reduce() algorithm: this was turned off but I think it will be worthwhile to have it in there in a limited set of cases. The current code that was in there will produce memory corruption so we can't have that, but if worked on I believe it could be valuable.

### LINKED ISSUE #677: Optimizer ability to copy instruction blocks
Optimizer ability to copy instruction blocks: we need to do this if we want to implement step 4 of our branch reduce steps

### LINKED ISSUE #681: Jump to ret hoisting inside of the register allocator
Jump to ret hoisting inside of the register allocator: there are many cases where we have something like:

```
.L8
movl $5, %eax
jmp .L10

.L9
movl $6, %eax
jmp .L10

.L10
ret
```

We should be able to perform a modified form of hoisting to move the ret into .L8/.L9 and then delete .L10 if it's got no more predecessors.

jackr276:
[Optimizer]: Update branch reduce to its final working state

Closes #674
Closes #677 
Closes #681

### DIFF
--- a/oc/compiler/postprocessor/postprocessor.c
+++ b/oc/compiler/postprocessor/postprocessor.c
@@ -513,7 +513,7 @@ static instruction_t* clone_instruction(instruction_t* source){
  * NOTE: This function will not delete anything. We are doing a straight copy from the source
  * to the destination *with the exception of any phi-functions*
  */
-static inline void copy_block(basic_block_t* destination, basic_block_t* source){
+static void copy_block(basic_block_t* destination, basic_block_t* source){
 	//Grab a cursor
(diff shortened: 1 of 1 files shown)


jackr276:
We will leave the branch hoisting code in for now as it may become useful sometime later down the road. It is working, it's just not producing results that seem desirable *at this time* but that could always change.

PR #683, opened 2026-03-03
[Instruction Selector]: Full support for floating point logical and/logical or

### LINKED ISSUE #568: Logical and/or for floating point values
Logical and/or for floating point values - this is a separate process from GP registers

jackr276:
[Instruction Selector]: Full support for floating point logical and/logical or

Closes #568

### DIFF
--- a/oc/test_files/float_logical_or_result_used.ol
+++ b/oc/test_files/float_logical_or_result_used.ol
@@ -1,0 +1,26 @@
+/**
+* Author: Jack Robbins
+* Test the interference of those pxor_clears we use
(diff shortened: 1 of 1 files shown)


jackr276:
There was an issue discovered when converting from ints to floats in chained logical or/and expressions. I suspect this issue will also appear inside of logical not/assigning from comparison operations as well. We are going to handle it in a separate issue tomorrow, as it is not in scope here. The test cases that had that crop up will be deactivated for this MR and then reactivated after our fix tomorrow.

PR #696, opened 2026-03-08
[Language Feature]: Ollie error system - callee-side implementation

### LINKED ISSUE #694: Ollie error subsystem
Ollie error subsystem: we need a way to represent errors in Ollie. It's not going to be anything like try-catching, or a result type, but instead we can pass errors out via some other register or stack region.

### LINKED ISSUE #695: Ollie error subsystem on the front-end
Ollie error subsystem on the front-end: we need to figure out how this will look on the front end. From function declarations to how the function call/raise is going to work to mandate that the user catches these errors.

### LINKED ISSUE #697: [Front-End]: Direct function pointer parsing with no params does not work
[Front-End]: Function pointer parsing with no params does not work:

When we parse a function pointer that has no parameters like:

```
let x:fn() -> i32 = my_func;
```

It fails to parse correctly. This is an error that needs to be fixed

### LINKED ISSUE #698: [Front-End]: Defining function pointers with no parameters does not work
[Front-End]: defining function pointers with no params does not work:

When we parse a function pointer that has no parameters like:

```
define fn() -> i32 as custom_function;
```

It fails to parse correctly. This is an error that needs to be fixed

### LINKED ISSUE #699: [Front-End]: "raises" clause should not allow duplicate errors
[Front-End]: "raises" clause should not allow duplicate errors - we'll need to add some safeguards for this in wherever we parse the raises clause to prevent the user from doing something like "raises (error1, error1)"

### LINKED ISSUE #704: [Front End]: Raise statement implementation in the parser
[Front End]: Raise statement implementation in the parser - we need the ability to do these raise statements inside of the parser. They are effectively return statements of a different kind

### LINKED ISSUE #705: [Front End]: Raise statement implementation inside of the CFG
[Front End]: Raise statement implementation inside of the CFG - we need CFG level support for handling these in a somewhat similar way to how ret statements are handled.

### LINKED ISSUE #706: [CFG]: Raise statements need to be treated as ret statements in validation
[CFG]: Raise statements need to be treated as ret statements in validation - when we do validations, we need to treat the raise statements as ret statements. This is especially true when we determine and insert return statements, ensuring that a function returns in every path.

### LINKED ISSUE #707: [CFG]: validate that defer statements do not get placed next to raise statements
[CFG]: validate that defer statements do not get placed next to raise statements: I think that the logic currently will do this so we need to check with a test case and guard against doing this.

### LINKED ISSUE #708: [Optimizer]: raise statement support
[Optimizer]: raise statement support - raise statements need to always be marked as valuable when mark and sweep runs in the optimizer.

### LINKED ISSUE #709: [Instruction Selector]: Normal returns on functions that can error must zero out %RDX
[Instruction Selector]: Normal returns on functions that can error must zero out %RDX - since %RDX is the error register, if we are returning from an errorable function normally we need to ensure that this has been zeroed out to avoid false positives for errors.

jackr276:
[Language Feature]: Ollie error system - initial implementation

Note that this is specifically dealing with the callee side of things. There is a whole extra task that involves the caller-side of ollie errors that we need to do, including the new handles statements.

Closes #694 
Closes #695 
Closes #697 
Closes #698 
Closes #699 
Closes #704 
Closes #705 
Closes #706 
Closes #707 
Closes #708 
Closes #709

### DIFF
--- a/oc/compiler/instruction/instruction.c
+++ b/oc/compiler/instruction/instruction.c
@@ -4233,7 +4233,7 @@ void print_instruction(FILE* fl, instruction_t* instruction, variable_printing_m
 		case RAISE_INSTRUCTION:
 			fprintf(fl, "ret /* --> raises error ");
 			print_variable(fl, instruction->source_register, mode);
-			fprintf(fl, " */");
+			fprintf(fl, " */\n");
 			break;
 
 		case NOP:
--- a/oc/test_files/error_raising_with_rdx_interference.ol
+++ b/oc/test_files/error_raising_with_rdx_interference.ol
@@ -1,0 +1,21 @@
+/**
+* Author: Jack Robbins
+* Test our ability to raise errors even if RDX is used
+*/
+
+pub fn! raise_with_rdx(x:i32, y:i32, z:i32) -> i32 {
+	if (x < 0) {
+		raise error;
+	}
+
+	if(z > 0) {
+		raise error;
+	}
+
+	ret x + y - z;
+}
+
+
+pub fn main() -> i32 {
+	ret 0;
+}


jackr276:
Memory check comes up clean:
image

PR #701, opened 2026-03-09
[Back-End]: Ret hoisting has been improved so that more than just ret statements will be hoisted up

### LINKED ISSUE #700: [Back-End]: Enhance ret hoisting to include stack deallocation statements
[Back-End]: Enhance ret hoisting to include stack deallocation statements. Currently ret hoisting only supports pure ret statements. In reality
we should fix this to include stack deallocation statements and possibly more so that we are even more efficient with our generated code.

jackr276:
[Back-End]: Ret hoisting has been improved so that more than just ret statements will be hoisted up

Closes #700

### DIFF
--- a/oc/compiler/postprocessor/postprocessor.c
+++ b/oc/compiler/postprocessor/postprocessor.c
@@ -17,7 +17,7 @@
  * would case more trouble than it's worth because it would
  * explode the binary size
  */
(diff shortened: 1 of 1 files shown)


jackr276:
Memory check is clean:
image

PR #711, opened 2026-03-13
[TRIAGED FIX]: fixes to div/mod/unsigned multiplication coalescence, and instruction selector constant folding

### LINKED ISSUE #710: TRIAGE: constant folding for division, multiplication is very brittle, not fully functional
TRIAGE: constant folding for division, multiplication is very brittle, not fully functional - we really need to take the time to rewrite this entirely. It does not make much sense and it is brittle to breakage, especially if we have values that are not powers of 2

### LINKED ISSUE #712: TRIAGE: unsigned integer multiply is generating extra move instructions
TRIAGE: unsigned integer multiply is generating extra move instructions

return_unsigned_multiply:
================== Stack Layout ===================
EMPTY
================== Stack Layout ===================
movl %esi, %eax
mull %edi /* Implicit Source: %eax -->  %eax */
movl %eax, %edi
movl %edi, %eax
ret /* --> %eax */

Notice the two extra moves. This is happening because we(mistakenly) do not have a temp assignee whenever we do unsigned multiplication. Just like signed division, we should have a temp assignee here and we *should not* be using op1 as the assignee.

### LINKED ISSUE #713: Left shift needs to be added to the inplace constant simplifier
Left shift needs to be added to the inplace constant simplifier: currently we do not support this but there is no good reason for us not to. It needs to be fully supported especially in light of our new optimizations

### LINKED ISSUE #714: Left/right shift constant arithmetic needs to be added
Left/right shift constant arithmetic needs to be added: currently we only support add, subtract and multiply. With the new optimization that is coming in we are going to need left/right shift as well.

jackr276:
[TRIAGED FIX]: fixes to div coalescence, and instruction selector constant folding

Closes #710 
Closes #712 
Closes #713 
Closes #714

### DIFF
--- a/oc/compiler/instruction_selector/instruction_selector.c
+++ b/oc/compiler/instruction_selector/instruction_selector.c
@@ -3161,33 +3161,6 @@ static u_int8_t simplify_window(instruction_window_t* window){
 
 
 	/**
-	 * There is a chance that we could be left with statements that assign to themselves
-	 * like this:
-	 *  t11 <- 2
-	 *
-	 *  Where t11 has no real usage at all. Since this is the case, we can eliminate the whole
-	 *  operation
-	 *
-	 *  These are guaranteed to be useless, so we can eliminate them
-	 */
-	if(window->instruction1 != NULL && window->instruction1->statement_type == THREE_ADDR_CODE_ASSN_CONST_STMT
-		//If we get here, we have a temp assignment who is completely useless, so we delete
-		&& window->instruction1->assignee->variable_type == VARIABLE_TYPE_TEMP
-		//Ensure that it's not being used at all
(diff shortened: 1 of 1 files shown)


jackr276:
Unsigned multiply has the exact same issue, we'll need to address this:

return_unsigned_multiply:
================== Stack Layout ===================
EMPTY
================== Stack Layout ===================
movl %esi, %eax
mull %edi /* Implicit Source: %eax -->  %eax */
movl %eax, %edi
movl %edi, %eax
ret /* --> %eax */

jackr276:
Also took the opportunity while being in the instruction selector to delete some rules that made no sense for us. These were all "redundant copy folders" that either were never hit or were already handled better by the register allocator's coalescence step.

PR #717, opened 2026-03-15
[Language Feature]: Ollie caller side error handling

### LINKED ISSUE #715: Account for error returns in caller-saved logic
Account for error returns in caller-saved logic: we need to account for the fact that the %rdx register will be populated alongside the %rdx register when we do error handling.

### LINKED ISSUE #716: Ollie "handle" statements
Ollie "handle" statements: these are essentially switch statements that will allow us to handle errors as we see fit. We first need to support this in the parser and then in the cfg, instruction selector, etc.

### LINKED ISSUE #718: Handle statements - parser side handling and error cases
Handle statements - parser side handling and error cases: we need test coverage for incorrect usage of handle statements, as well as parser-side translation of handle statements into the AST.

### LINKED ISSUE #719: Testing for all fail cases related to error handling
Testing for all fail cases related to error handling: we need extensive test cases for all error handling, especially the fail cases. We want to catch bugs now when we're building this system from the ground up.

### LINKED ISSUE #720: Compiler should throw an error if a raises statement contains an error that is not actually raised by the function
Compiler should throw an error if a raises statement contains an error that is not actually raised by the function

### LINKED ISSUE #721: CFG implementation of handle statements
The handle clause itself will be translated internally into a switch statement. We need CFG level handling for this if and only if we see it in the function itself

### LINKED ISSUE #722: Error handling with void returning functions
Error handling with void returning functions: we need to validate/add special handling for this case. We also need to add fail case testing for this on the parser side

### LINKED ISSUE #724: Ability to effectively swallow an error using the "ignore" keyword
Ability to effectively swallow an error using the "ignore" keyword: we need the ability to just swallow an error and go no further with it. We will add the special "ignore" keyword to account for that.

jackr276:
[Language Feature]: Ollie caller side error handling

Closes #715 
Closes #716 
Closes #718 
Closes #719 
Closes #720 
Closes #721 
Closes #722 
Closes #724

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -2949,6 +2949,9 @@ static cfg_result_package_t emit_return(basic_block_t* basic_block, generic_ast_
 	//Once it's been emitted, we'll add it in as a statement
 	add_statement(current, ret_stmt);
 
+	//This is always a predecessor of the function exit statement
+	add_successor(current, function_exit_block);
+
 	//Give back the results
 	return return_package;
 }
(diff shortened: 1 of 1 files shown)


jackr276:
We are ready to begin validations

jackr276:
Everything here is looking how we would expect. There will certainly be more that arises from this but for right now we are in a good spot to merge.

PR #726, opened 2026-03-18
[Parser]: handles statements now properly coerce constants when they're given as the result to an error handler

### LINKED ISSUE #725: Moving from integer constants into floats does not work
Moving from integer constants into floats does not work:

Something like float t7 <- \<i8\>2 is not working and currently throws an internal compiler error

jackr276:
[Parser]: handles statements now properly coerce constants when they're given as the result to an error handler

Closes #725


jackr276:
This issue turned out to be simpler than expected. We are ready to now start reviewing.

jackr276:
We can call this bug fixed, good for us to merge

PR #732, opened 2026-03-21
[Language Feature]: Ollie elaborative stack param

### LINKED ISSUE #729: Elaborative param support inside of the function pointer types & appropriate test coverage
Elaborative param support inside of the function pointer types & appropriate test coverage: we need to ensure that this is fully supported including when we assign for function types, and when we have function pointers/indirect calls.

### LINKED ISSUE #730: Validation to ensure elaborative params are always the last listed parameter
Validation to ensure elaborative params are always the last listed parameter: we need test cases for fail scenario validation

### LINKED ISSUE #731: Inlined functions may not have stack params
Inlined functions may not have stack params: new decision but one I think is on the right track - we should not allow inlined functions to have stack params of any kind(this includes elaborative params)

### LINKED ISSUE #733: Parsing of elaborative param in inline function pointer definition
Parsing of elaborative param in inline function pointer definition: we need to do this in the parser and add test coverage for all over our fail cases

### LINKED ISSUE #734: Parsing of elaborative params inside of direct function calls
Parsing of elaborative params inside of direct function calls: we need to be able to parse these. Remember the rules: an elaborative param means that we can see: 0, 1 or many of the provided/a compatible elaborated type

### LINKED ISSUE #735: Dynmaically sized stacks to enable different stack sizes upon each call site of an elaborative param function
Dynmaically sized stacks to enable different stack sizes upon each call site of an elaborative param function

### LINKED ISSUE #736: Paramcount statement implementation inside of the CFG
Paramcount statement implementation inside of the CFG: this should be relatively simple to implement - all that a paramcount is going to need to do is grab the first 4 bytes from the variable passed inside of it.

### LINKED ISSUE #737: Dynamic stack implementation in the CFG/register allocator for function calls with elaborative params
Dynamic stack implementation in the CFG/register allocator for function calls with elaborative params: we know that there will always be some kind of stack presence here so we'll need to create/emit a unique stack for every function call with an elaborative param

### LINKED ISSUE #738: Implementation of array access for elaborative param variables
Implementation of array access for elaborative param variables: this will be a CFG enhancement and we'll need to go through and check to see if our base address is a params variable. If it is, then we're going to need to account for the base address and the adjustment of 4 bytes which is where the size is stored.

### LINKED ISSUE #740: Code Quality: remove conditional reassignment of final block values
Code Quality: remove conditional reassignment of final block values:

We need to remove patterns like this: 

Image

Instead we should just always be reassigning.

### LINKED ISSUE #741: [Front End]: indirect function calls with elaborative stack param implementation
[Front End]: indirect function calls with elaborative stack param implementation - this issue is just opened for tracking on that

### LINKED ISSUE #742: [Front End]: Direct function calls with elaborative stack implementation
[Front End]: Direct function calls with elaborative stack implementation - this issue exists just for tracking

### LINKED ISSUE #743: Constant assignment in function calls with stack params generating excessive interference
This is causing excessive interference due to the way we currently process. We need to fix this for constants - perhaps even maintaining separate lists. All that I currently know is that this is generating ~1.5x the instruction count that it should

```
leaq elaborative_param(%rip), %rdx
movl $1, %edi
movl $2, %eax
movl $3, %ecx
push %rdx
subq $16, %rsp
movl $2, (%rsp)
movl %eax, 4(%rsp)
movl %ecx, 8(%rsp)
call *%rdx /* --> %eax */
addq $16, %rsp
pop %rdx
movl %eax, %ecx
movl $1, %edi
movl $4, %r9d
movl $5, %r10d
movl $6, %r8d
movl $7, %esi
movl $8, %eax
push %rcx
subq $32, %rsp
movl $5, (%rsp)
movl %r9d, 4(%rsp)
movl %r10d, 8(%rsp)
movl %r8d, 12(%rsp)
movl %esi, 16(%rsp)
movl %eax, 20(%rsp)
call *%rdx /* --> %eax */
addq $32, %rsp
pop %rcx
addl %eax, %ecx
movl %ecx, %eax
ret /* --> %eax */
```

### LINKED ISSUE #744: Update the indirect call to use the new parameter_result_array
Update the indirect call to use the new parameter_result_array: this should result in much less interference when we have stack passed params that are storing constants.

### LINKED ISSUE #745: Update the direct call to use the new parameter results array
Update the direct call to use the new parameter results array: this should provide the same boost that we've seen it provide to the indirect call when it comes to redundant assignments/interference

jackr276:
[Language Feature]: Ollie elaborative stack param: ollie has been enhanced to offer an elaborative stack param using the "params" and "paramcount" keywords. These new keywords allow the user to pass parameters a variable number of parameters.

Closes #729
Closes #730
Closes #731 
Closes #733 
Closes #734 
Closes #735 
Closes #736 
Closes #737 
Closes #738 
Closes #740 
Closes #741
Closes #742
Closes #743
Closes #744  
Closes #745

### DIFF
--- a/oc/test_files/elaborative_param_with_callee_stack.ol
+++ b/oc/test_files/elaborative_param_with_callee_stack.ol
@@ -1,0 +1,26 @@
+/**
+* Author: Jack Robbins
+* Test the case where we have an elaborative param that
+* also has local stack allocations
+*/
+
+
+pub fn elaborative_params(x:mut i32, y:params i32) -> i32 {
+	let arr:i32[] = [1, 2, 3, 4];
+
+	for(let i:mut i32 = 0; i < paramcount(y); i++){
+		x += y[i];
+
+		if(i < 4) {
+			x += arr[i];
+		}
+	}
+
+	ret x;
+}
+
+
+pub fn main() -> i32 {
+	//Valide use where we don't pass anything
+	ret @elaborative_params(1);
+}


jackr276:
Memory Check comes up clean:
image

jackr276:
Final case here validated:
elaborative_param_with_callee_stack.ol

PR #760, opened 2026-03-29
[Utility]: "--to-object-file" flag added to skip linking entirely

### LINKED ISSUE #752: Optional flag so that OC can generate .o object files
This does technically violate our philosophy of only having one file to build everything, but we need to be able to output assembled(but pre-linker) .o files just in case they are needed. Thinking that the flag should be something like "-c".

jackr276:
[Utility]: "--to-object-file" flag added to skip linking entirely. Note that this feature should only ever be used if you specifically need the object file only, which is a very rare case.

Close #752


jackr276:
Memory check comes up clean:
image

jackr276:
No test results/generated assembly has changed with this update. The .o file output was verified manually as there is no test coverage for it currently(nor will there likely ever be)

PR #768, opened 2026-04-01
[Language Feature]: Struct to struct, union to union copying implemented in the let/assign(=) cases

### LINKED ISSUE #764: Struct copy assignment - parser side support
Struct copy assignment - parser side support: we need the parser to recognize struct assignment to another struct as a valid operation and generate the appropriate AST nodes for it

### LINKED ISSUE #765: Struct to struct assignment via the type system
Struct to struct assignment via the type system: we need to allow for this to happen. Currently it is only partially supported by "types_assignable" I believe.

### LINKED ISSUE #766: Nested struct copying
Nested struct copying: we need to verify that nested struct copying works. This refers to copying where we have a nested inner struct and we need to clone into that. This will involve us messing with the postfix expression system to get it to avoid derferencing.

### LINKED ISSUE #767: Copying with unions should work as well
Copying with unions should work as well: pretty self explanatory - we need to be able to do the same thing with unions.

### LINKED ISSUE #769: [TRIAGE]: Super small structs(4 bytes and under) are not at least 4-byte aligned
Something like a struct of 3 chars, even though having it 3 bytes large follows the "multiple of the largest member size" rule, we need these structs to be at least 4 bytes.

### LINKED ISSUE #770: Instruction selector support for new double quad word type
Instruction selector support for new double quad word type: the double quad word type is really only seen in move instructions. We use movdqu/movdqa when we are loading and then movaps when we are storing. This is because we can guarantee alignment when storing.

### LINKED ISSUE #771: Ability to test for alignment when doing 16 byte loads
Ability to test for alignment when doing 16 byte loads: when we are doing 16 byte loads, we need to be able to determine if the memory region can be guaranteed to be aligned or not. Memory regions are only guaranteed alignment if they come from the local stack region. Anything else - it is not at all guaranteed.

jackr276:
[Language Feature]: Struct to struct, union to union copying implemented in the let/assign(=) cases

Closes #764 
Closes #765 
Closes #766 
Closes #767 
Closes #769 
Closes #770
Closes #771

### DIFF
--- a/oc/compiler/instruction_selector/instruction_selector.c
+++ b/oc/compiler/instruction_selector/instruction_selector.c
@@ -1328,8 +1328,8 @@ static void convert_memory_copy_statement_into_loads_and_stores(instruction_wind
 
 	//Maintain the current offset. This is going to be the same for the source and destination
 	u_int64_t current_offset = 0;
-	//Always use the source to get how much we want to copy
-	u_int64_t remaining_copy_amount = source_memory_address_var->type->type_size;
+	//We always use the dedicated field to determine how many bytes we should be copying
+	u_int64_t remaining_copy_amount = memory_copy_statement->optional_storage.byte_amount_to_copy;
 
 	//We need to keep track of the last instruction. This will be constantly updated by every function we call
 	instruction_t* last_instruction = memory_copy_statement;


jackr276:
Overall good to merge

PR #774, opened 2026-04-03
[TRIAGED FIX]: Memory alignment should now allow for aligned SIMD instructions for struct copying

### LINKED ISSUE #772: [TRIAGE]: alignment is not 16 byte aligned on function entry
[TRIAGE]: alignment is not 16 byte aligned on function entry - all alignments need to be 16-byte aligned so that we are able to use the movaps/movapd/movdqa assembly instructions. These instructions will segfault if a non-aligned memory segment is attempted on.

### LINKED ISSUE #773: [TRIAGE]: Require that struct addresses be aligned to start at 16 bytes and end at 16 byte aligned addresses
[TRIAGE]: Require that struct addresses be aligned to start at 16 bytes and end at 16 byte aligned addresses. This is because *if* we decide that we want to copy a struct/union, we need to be able to use the movdqa/movapd/movaps. instructions.

### LINKED ISSUE #775: Caller/callee saving must also always guarantee 16 byte aligned chunks
Caller/callee saving must also always guarantee 16 byte aligned chunks: if we are doing caller/callee saving, we need to ensure that the amount that we are caller/callee saving by is always even. This means that we can't push an odd number of registers if we're only pushing for example.

jackr276:
[TRIAGED FIX]: Memory alignment should now allow for aligned SIMD instructions for struct copying

Closes #772 
Closes #773 
Closes #775

### DIFF
--- a/oc/compiler/register_allocator/register_allocator.c
+++ b/oc/compiler/register_allocator/register_allocator.c
@@ -30,9 +30,11 @@ const general_purpose_register_t gen_purpose_parameter_registers[] = {RDI, RSI, 
 const sse_register_t sse_parameter_registers[] = {XMM0, XMM1, XMM2, XMM3, XMM4, XMM5, XMM6, XMM7};
 
 /**
- * Our dummy push register(if one is needed) will always be R12
+ * Our dummy push register(if one is needed) will always be R12 for caller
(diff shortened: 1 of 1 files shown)


jackr276:
Memory check comes up clean:
image

PR #777, opened 2026-04-03
[TRIAGED FIX]: Function calls now obey the System V ABI alignment mandate when it is necessary

### LINKED ISSUE #776: [TRIAGE]: System V ABI guarantees for call instructions
[TRIAGE]: System V ABI guarantees for call instructions - before we make *any* function call instruction, we must guarantee that the stack
is 16 byte aligned. This follows through with our other core assumption that whenever we enter into a function, the stack is 8, but not 16, byte aligned.

### LINKED ISSUE #778: [Front End]: Transitive closure discovery of functions that need to be aligned
[Front End]: Transitive closure discovery of functions that need to be aligned - we need to use the transitive closure so that we are only inserting stack alignment functions on paths where they are necessary. This should be simple with the transitive closure of the function call graph, we will go through and flag every function that can reach a default flagged function that needs to be aligned. The entire chain of functions need to be aligned in order for this to work.

jackr276:
[TRIAGED FIX]: Function calls now obey the System V ABI alignment mandate when it is necessary

Closes #776 
Closes #778

### DIFF
--- a/oc/compiler/compiler.c
+++ b/oc/compiler/compiler.c
@@ -25,8 +25,8 @@
 #include "utils/error_management.h"
 
 //The number of errors and warnings
-u_int32_t num_errors;
-u_int32_t num_warnings;
+u_int32_t num_errors = 0;
+u_int32_t num_warnings = 0;
 
 //Objectfile opt for getopt_long
 #define objectfile_opt 5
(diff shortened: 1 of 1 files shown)


jackr276:
Changes looking good, the memory_copy_structs that was segfaulting now works.

PR #787, opened 2026-04-05
[Language Feature]: Support for Ollie namespaces and namespace references

### LINKED ISSUE #761: Ollie namespaces
Ollie namespaces and namespace references: we will be adding namespaces to Ollie. This will mean that the function, type and variable symtabs will need to be updated to support this concept

### LINKED ISSUE #762: Ollie namespace references
Ollie namespace references: we need a way to uniquely identify functions inside of a namespace. If no reference is found, then we look into a higher namespace(same concept with variables and scopes). We can use the "::" as a namespace qualifier

### LINKED ISSUE #784: Lexer side support for the namespace keyword
Lexer side support for the namespace keyword is required

### LINKED ISSUE #785: [Parser]: screening is needed so that the main function may not be declared inside of a namespace
[Parser]: screening is needed so that the main function may not be declared inside of a namespace - we can't have the main function inside
of a namespace because that would cause the name to be mangled, making the program completely "unrunnable"

### LINKED ISSUE #786: [Symtab]: enhance the function symtab to support sheafs, namespaces in a similar way that the variable symtab supports scopes
[Symtab]: enhance the function symtab to support sheafs, namespaces in a similar way that the variable symtab supports scopes

### LINKED ISSUE #788: Function symtab should be more like an n-ary tree, where we have children arrays and parent references
Function symtab should be more like an n-ary tree, where we have children arrays and parent references. This is going to be done to support
namespaces like namespace1::child1 and namespace1::child2 and maintain tracking for their relationship.

### LINKED ISSUE #789: Ability to parse fully qualified function names
Ability to parse fully qualified function names: we need the ability to parse fully qualified function names both at the function pointer end and at the point of a function call. This involves running through the :: operator until we hit the end. Extensive testing for this will be required.

### LINKED ISSUE #793: [Front End]: Update the parser to allow for predeclared functions inside of a namespace
[Front End]: Update the parser to allow for predeclared functions inside of a namespace: we will allow for function predeclarations but we will
not allow for any global variables.

### LINKED ISSUE #794: [Front End]: Namespace name mangling pass is needed to guarantee unique name generation
[Front End]: Namespace name mangling pass is needed to guarantee unique name generation - we need to have a pass at the end of the parser that runs through all functions and mangles their names using the namespace that they're in. This will guarantee that everything has a unique name in the final assembly.

### LINKED ISSUE #795: [Parser]: Rewrite the primary_expression parser to not generate rely on the identifier() function
[Parser]: Rewrite the primary_expression parser to not generate rely on the identifier() function - there's really no reason for it to do this and it won't work with what we need to do for fully qualified names anyway. We can refactor to not rely on it.

### LINKED ISSUE #799: Public function enforcement in namespaces
Public function enforcement in namespaces: you cannot call/grab pointers to functions not labeled as "pub" in a given namespace *if* you are not also in that namespace. Functions in higher namespaces are publicly visible to every descendant namespace regardless of the keyword. This is all enforced at the parser level. We require extensive testing for all of these scenarios.

### LINKED ISSUE #800: Add success cases for finding a function inside of a predecessor namespace
Add success cases for finding a function inside of a predecessor namespace: currently we only have fail cases which means that this is incomplete

### LINKED ISSUE #801: Test cases for function pointers in namespaces are needed
Test cases for function pointers in namespaces are needed: we need the same battery of tests that was done for regular function calls to ensure that the access control is working properly

jackr276:
[Language Feature]: Support for Ollie namespaces and namespace references

Closes #761 
Closes #762 
Closes #784 
Closes #785 
Closes #786 
Closes #788 
Closes #789
Closes #793 
Closes #794 
Closes #795 
Closes #799
Closes #800 
Closes #801


jackr276:
Memory check comes up clean:
image

jackr276:
Validated:
basic_namespace_test.ol(validated via run)
invalid_attempt_to_access_private_namespace_function.ol 
invalid_attempt_to_get_private_function_pointer.ol
invalid_duplicate_functions_in_namespace.ol
invalid_duplicate_namespace.ol
invalid_duplicate_predeclared_functions_in_namespace.ol
invalid_function_not_in_namespace.ol
invalid_main_function_in_namespace.ol
invalid_namespace_not_in_namespace.ol
invalid_qualified_function_lookup.ol
invalid_qualified_function_pointer.ol
invalid_unrelated_namespace_non_qualified_call.ol
namespace_function_pointer.ol(validated via run)
valid_call_to_function_in_higher_namespace.ol(validated via run)
valid_duplicate_function_name_test.ol(validated via run)
valid_private_function_pointer_predecessor_namespace.ol(validated via run)
valid_private_function_predecessor_namespace.ol(validated via run)

jackr276:
This is done. There are concerns with the current implementation of the heap_queue_t. We currently have it using a node implementation. We're going to need to make this a circular queue as we can't afford the allocation overhead of the current implementation.

PR #792, opened 2026-04-06
[Instruction Selector]: Update alignment determination strategy to give accurate alignment guarantees

### LINKED ISSUE #790: Struct alignment must always be 16 bytes if we intend on using the movaps struct copy
Struct alignment must always be 16 bytes if we intend on using the movaps struct copy: this is currently not being obeyed for a given array of structs inside of "struct_in_array_copy_assignment.ol" and causes a segfault.

jackr276:
[Instruction Selector]: Update alignment determination strategy to give accurate determinations

Closes #790 
Closes #791


jackr276:
Validated:
struct_assignment_nested.ol(validated with stdout)
struct_assignment_nested_rhs.ol(validated with stdout)
struct_in_array_copy_assignment.ol(validated with stdout)

PR #798, opened 2026-04-09
[Type System]: Array types now properly show the number of members when possible

### LINKED ISSUE #796: Array types need to be able to print out the number of members
Array types need to be able to print out the number of members: we should be able to print this out from inside of the type system. This should be stored inside of the brackets.

### LINKED ISSUE #797: [Utils]: New "dynamic_string_insert" function
[Utils]: New "dynamic_string_insert" function: we should have a helper that allows us to insert into a dynamic string. We will need this to deal with the array bounds insertion. This should get coverage inside of the dynamic string test.

jackr276:
[Type System]: Array types now properly show the number of members

Closes #796 
Closes #797

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -447,6 +447,9 @@ selector_test_non_timed: instruction_selector_test
 string_test: dynamic_string_test
 	$(OUT_LOCAL)/dynamic_string_test
 
+string_test_debug: dynamic_string_testd
+	$(OUT_LOCAL)/dynamic_string_testd
+
 compiler_test: oc
 	@for input in $(inputs); do \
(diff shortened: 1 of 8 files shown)


jackr276:
The array type member implementation has happened. We may still want to implement generic array types for parameters/return types(still unsure about this though). As for right now this enhancement is done and ready to be verified.

jackr276:
Memory check comes up clean:
image

jackr276:
Verification was successful. There were no changes to generated code as part of this. The only thing that changed are various error messages. As of now this enhancement is considered done. Any other changes are going to be considered as additional enhancements.

jackr276:
All looks good, merging

PR #804, opened 2026-04-11
[Compiler Utilities]: The heap_queue_t has been rewritten to use an underlying circular queue

### LINKED ISSUE #802: [Utilities]: Rewrite the heap queue into a circular queue
[Utilities]: Rewrite the heap queue into a circular queue: currently the heap queue has an underlying node structure. This is inefficient as we need to dynamically allocate each node for one piece of data. This also means that the queue is not easily reusable. We need to fix this by rewriting the heap queue to use an underlying array with a circular queue design.

### LINKED ISSUE #803: [Utility]: Test suite update after the queue has been rewritten as a circular queue
[Utility]: Test suite update after the queue has been rewritten as a circular queue: creating this generic issue for any testing update that we will need(I'm assuming that there will be some)

jackr276:
[Compiler Utilities]: The heap_queue_t has been rewritten to use an underlying circular queue. This greatly reduces our amount of memory allocation and allows for us to operate more efficiently. It probably should have just been a circular queue from the start.

Closes #802 
Closes #803

### DIFF
--- a/oc/compiler/utils/queue/heap_queue.c
+++ b/oc/compiler/utils/queue/heap_queue.c
@@ -13,6 +13,7 @@
 #include "heap_queue.h"
 #include <stdlib.h>
 #include <stdio.h>
+#include <string.h>
 #include <sys/types.h>
 #include "../constants.h"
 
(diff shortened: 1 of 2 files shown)


jackr276:
The dynamic resize has been confirmed to work by lowering the default size

jackr276:
Memory check comes up clean:
image

jackr276:
There are no changes between the new output and the old. This is what was expected as the underlying queue should behave in the exact same way, albeit more efficiently. This is ready to merge.

PR #808, opened 2026-04-12
[Optimizer]: Base case implementation of global value numbering

### LINKED ISSUE #805: [Optimizer]: Global value numbering pass and optimization
[Optimizer]: Global value numbering pass and optimization. We will use global value numbering to detect expressions that end up calculating the exact same thing. This will be done, I think, as part of the optimizer but that may change.

### LINKED ISSUE #806: [Compiler Utility]: Value numbering general usage hash map
[Compiler Utility]: Value numbering general usage hash map: we need the ability to create a general purpose hashmap. Currently we do not have one of these so we'd need to make it for value numbering.

### LINKED ISSUE #807: Need a way to handle value numbering if we have two variables with the same names at different scopes
Need a way to handle value numbering if we have two variables with the same names at different scopes: this is a potential bug that I can foresee where say we have two variables called "x_1" but they're just in different lexical scopes inside of the function. This could lead to collisions and an incorrect value numbering. Potential fix(not proven) is to prepend the lexical scope to avoid confusion(something like 1_x1 and 2_x1).

### LINKED ISSUE #809: [Value Numberer]: Helper in order to generate the value name of a variable is needed
[Value Numberer]: Helper in order to generate the value name of a variable is needed. It is crucial that this is consistent so that we are able to actually look things up using these textual strings when the time comes.

### LINKED ISSUE #826: [Value Numberer]: Ability to exclude certain instructions if we can't optimize them
[Value Numberer]: Ability to exclude certain instructions if we can't optimize them - some instructions like comparison instructions are going to be used by branch statements. We currently have a crude solution in place where we just don't consider them. This should be updated if possible and we should be able to detect if we are relying on an instruction setting condition codes or not.

### LINKED ISSUE #827: [Instruction Simplifier]: Mod by power of 2 is breaking the one assignment per temp rule
We are generating code like this:

t26 <- ^t1_0
t26 <- t26 >> 31
t26 <- t26 >> 30

This is bad. We can't be having this happen as it will confuse the value numberer and it's just fundamentally wrong for OIR. This should
instead be something like 

t26 <- ^t1_0 >> 31
t27 <- t26 >> 30

This is now perfectly adequately handled. I believe the way that this currently is is a remnant of the older worse IR.

### LINKED ISSUE #828: [Value Numberer]: Implementation of the value name substitution for GVN
[Value Numberer]: Implementation of the value name substitution for GVN

### LINKED ISSUE #829: [Value Numberer]: Ability to run mark and sweep after we do value name substitutions
[Value Numberer]: Ability to run mark and sweep after we do value name substitutions. We need to do this to completely remove dead code from the equation that gets generated by accident. We should also intelligently decide whether or not a pass like this has happened to avoid extra work.

### LINKED ISSUE #830: [Instruction Selector]: Bring a simplified version of mark and sweep to the selector
[Instruction Selector]: Bring a simplified version of mark and sweep to the selector - I think that this best for us to instead of making an API that, so that we can control for any specific behaviors that we're after

### LINKED ISSUE #831: [Value Numberer]: phi function simplification in the numberer
[Value Numberer]: phi function simplification in the numberer: we need the phi function simplifier, as well as a redundant phi
function remover, in the value numberer.

### LINKED ISSUE #832: [General]: Completely remove useless "is_stack_pointer" flag
[General]: Completely remove useless "is_stack_pointer" flag: this was noticed in the three_addr_var_t struct. I'm not sure why it's there but we should just delete it.

### LINKED ISSUE #833: [Value Numberer]: Phi function simplification test case
[Value Numberer]: Phi function simplification test case - we need a test case to determine if this is actually working or not.

### LINKED ISSUE #834: [Instruction Simplifier]: Account for areas where a temp var is used more than once
[Instruction Simplifier]: Account for areas where a temp var is used more than once - this can happen now because of how the value numberer works. Usually this isn't a problem but it is an issue if we're dealing with binary operations that "think" that the temp var is only used once.

jackr276:
[Optimizer]: Base case implementation of global value numbering

Closes #805 
Closes #806 
Closes #807 
Closes #809 
Closes #826 
Closes #827 
Closes #828 
Closes #829 
Closes #830 
Closes #831 
Closes #832
Closes #833 
Closes #834


jackr276:
The basics here are down. We are going to verify and merge and then continue to build upon this base

jackr276:
Memory check comes up clean:
image

jackr276:
Validated:
array_testing.ol
function_call_stack_params_caller_saving.ol
indirect_function_call_stack_params_caller_saving.ol
macro_parameter_multiple_use.ol(known bug, new issue will be opened)

now we have a bug to resolve - we're going to put this on hold until that bug is resolved. Taking a break for tonight, bug will be resolved come morning.

jackr276:
We are ready to begin revalidating from where we left off. Nothing that we've previously validated changed luckily so we don't need to redo that.

jackr276:
Validate:
macro_parameter_multiple_use.ol
modulo-by-power-of-2.ol(validated via run)
parameter_passing.ol
reg_alloc_stress_test.ol
stack_passed_array_param.ol
value_numberer_array_arithmetic.ol 
value_numberer_stress_test.ol
value_numbering_conditionals.ol(validated via run)
value_numbering_phi_simplification.ol

jackr276:
We are finally good to go here. Remember that this is a base case, and the intention is that we will be adding more to the value numberer over time.

PR #825, opened 2026-04-18
[Parser]: parser can now recognize and reorder eligible expressions to guarantee that the constant is in the right hand side always

### LINKED ISSUE #824: [Parser]: Ability to reorder commutative arithmetic expressions
[Parser]: Ability to reorder commutative arithmetic expressions: we need the ability to, in the parser, reorder commutative expressions if it serves us. We always prefer to have the constant(if there is one) in the second position in the expression. So for example, if we are computing 5 + x, we should reorder that to x + 5. We will do this so long as the type and operator allow.

jackr276:
[Parser]: parser can now recognize and reoreder eligible expressions to guarantee that the constant is in the right hand side always

Closes #824

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -210,6 +210,15 @@ static inline u_int8_t is_copy_assignment_required(generic_type_t* destination_t
  */
 static inline u_int8_t is_type_commutative_for_operation(generic_type_t* type, ollie_token_t op){
 	switch(op){
+		/**
+		 * For multiplication, any type that is *not* a float is commutative
+		 */
+		case STAR:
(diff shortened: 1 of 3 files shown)


jackr276:
Memory check comes up clean:
image

jackr276:
Validated:
caller_saving.ol
handle_statement_in_use.ol
indirect_caller_saving.ol
lex_test6.ol
pointer_deref.ol
reorder_bitwise.ol
reordering_test.ol
valid_private_function_pointer_predecessor_namespace.ol
valid_private_function_predecessor_namespace.ol

jackr276:
This is looking good. The reordering is one more step towards making the IR very reliable for optimizations, especially value numbering.

PR #855, opened 2026-04-28
[CFG]: Refactor if-else-if-else handling in the CFG to avoid extra block allocation

### LINKED ISSUE #853: [CFG]: fix the if-else-if handler and the extra block allocations that it causes
[CFG]: fix the if-else-if handler and the extra block allocations that it causes: our current pattern causes extra basic block allocations. We need to fix this pattern so that else statements are not causing us to allocate blocks unnecessarily. It's not a big problem now but it will be for larger programs.

### LINKED ISSUE #854: [CFG]: Completely remove the is_conditional flag from emit_expression
[CFG]: Completely remove the is_conditional flag from emit_expression - this is an artifact of the way that we used to do things and is now completely useless

### LINKED ISSUE #856: [CFG]: make ret assignments optional if we have temporary variables, nonconverting moves
[CFG]: make ret assignments optional if we have temporary variables, nonconverting moves - currently we are doing way too much assignment at the ret end. We really only need an assignment if we're converting between types or moving out of a temp var. We need to add handling for this in the CFG.

### LINKED ISSUE #858: [CFG]: Remove unnecessary copy assignment for assignment statements
[CFG]: Remove unnecessary copy assignment for assignment statements - this enhancement was previously done for let statements, we should be doing the same for assignment statements.

### LINKED ISSUE #859: [CFG]: ensure all copy assignment removal is proper
[CFG]: ensure all copy assignment removal is proper - proper in this case means that we are checking for variable equality and that the variable that we're removing is a temp var. Failure to do this will create weird results.

jackr276:
[CFG]: Refactor if-else-if-else handling in the CFG to avoid extra block allocation

Closes #853 
Closes #854
Closes #856 
Closes #858 
Closes #859


jackr276:
This is looking good now, we are ready to start reviewing.

jackr276:
Memory check is clean:
image

jackr276:
Validated:
basic_declare_mutability_invalid.ol
basic_negation_instruction.ol
binary_operation_identical_operands_branching.ol
break_in_if.ol
c_switch_nested_in_switch.ol
case_statements.ol
continue_in_if.ol
elaborative_param_variable_entries.ol
elaborative_param_variable_entries_indirect.ol
empty_ifs.ol
float_logical_not_branch.ol
floating_point_branching.ol
if_else_if_tester.ol
indirect_caller_saving.ol
invalid_attempt_assign_to_elaborative_param.ol
invalid_raise_non_error_type.ol
invalid_raises_in_unerrorable_function.ol
logical_not_short_circuit.ol
mutate_after_init.ol
negate_indirect.ol
negative_in_switch.ol
nested_c_switches.ol
nested_switch.ol
nested_switch_default.ol
parameter_passing.ol
reassign_immut_pointer.ol
reg_alloc_stress_test.ol 
ret-hoisting-with-stack.ol
return_through_if_else_if.ol
short_circuit.ol
short_circuit_function_call.ol
short_circuit_var_only.ol
short_circuit_var_only_floats.ol
short_circuit_with_ternary.ol
ssa_test10.ol
ssa_test7.ol
ssa_test8.ol 
structures.ol
switch_always_true_optimization.ol
test_prog1.ol 
test_prog2.ol
test_prog7.ol
testing_void_declaration.ol

jackr276:
We are looking good here, all ready to merge.

PR #867, opened 2026-05-02
[FULL STACK]: Ability to write and produce infinite loops

### LINKED ISSUE #866: [CFG]: Block merging is done too freely, needs restrictions
[CFG]: Block merging is done too freely, needs restrictions - currently we allow things like merging the function entry block with loop entry statements. This screws everything up and as such should not be allowed. We need to put restrictions on the merge_blocks system.

### LINKED ISSUE #868: [CFG]: Ollie infinite loop support
[CFG]: Ollie infinite loop support - currently we know that infinite looping breaks *at least* the postdominator algorithm, probably more things as well. We need to fix this.

jackr276:
[FULL STACK]: Ability to write and produce infinite loops

Closes #866 
Closes #868

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -100,6 +100,14 @@ typedef enum{
 } emit_dominance_frontier_selection_t;
 
 
+//Enum for branch conditional types
+typedef enum {
+	BRANCH_CONDITIONAL_UNKNOWN,
+	BRANCH_CONDITIONAL_ALWAYS_TRUE,
+	BRANCH_CONDITIONAL_ALWAYS_FALSE
+} branch_conditional_truthfullness_t;
+
+
 /**
  * An enum for declare and let statements letting us know what kind of variable
  * that we have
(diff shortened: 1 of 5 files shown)


jackr276:
Our (maybe temporary) solution to this problem is to introduce virtual edges whenever an infinite loop is detected. Currently this is only being done in the CFG but it may be expanded later on. I say temporary because I do not know if this will still be necessary once we move to using strongly connected components(SCC) for dominator analysis. However with our current scheme this is absolutely necessary. An interesting side effect is that in the optimizer, we will now need to mark jump statements as important if they have a virtual edge on them. This makes the optimizer mark branches correctly. This also may need to be removed once we are done with the enhancement to our dominator analysis.

jackr276:
Memory check is clean:
image

jackr276:
There are no changes to existing files. The only thing that's changed are the two new testers.

jackr276:
Validated:
infinite_loop.ol
virtual_exit_do_while.ol

jackr276:
We are good to go here(for now with the infinite loops), merging

PR #906, opened 2026-05-15
[Compiler Utilties]: Ollie Run Validator test utility

### LINKED ISSUE #901: [Compiler Utilities]: New test suite that allows us to "validate via run"
[Compiler Utilities]: New test suite that allows us to "validate via run" - we have been doing this manually and there's really no reason why we should be. In the same way the memory checker is automated we should be automating this too.

### LINKED ISSUE #902: [Compiler Utilities]: New pseudo-syntax for ollie unit tests that allows us to know that a test is run validation eligible
[Compiler Utilities]: New pseudo-syntax for ollie unit tests that allows us to know that a test is run validation eligible. Some kind of tag like [OUNIT] or something like that

### LINKED ISSUE #903: [Compiler Utilities]: Validate via run pseudo syntax implementation
[Compiler Utilities]: Validate via run pseudo syntax implementation - we need to devise what it is and then actually implement it

### LINKED ISSUE #904: [Compilier Utilities]: Validate via run tie in to overall CI infrastructure
[Compilier Utilities]: Validate via run tie in to overall CI infrastructure - this includes the needed makefile updates to get this to run on the Github server.

### LINKED ISSUE #905: [Compiler Utilities]: Convert the memory checker to use the same array division strategy
[Compiler Utilities]: Convert the memory checker to use the same array division strategy - this will help us avoid having to use a mutex to lock the input array/array index and just makes more sense overall.

jackr276:
[Compiler Utilties]: Ollie Run Validator test utility

The Ollie Run Validator allows use to do automated "Validated Via Run". This is going to be an enormous time saver because the developer no longer has to validate via run manually. Going forward any/all new test suites that the developer wished to validate via console output(a good practice) can just use "OUNIT: [console = <return value>]. 

Closes #901 
Closes #902 
Closes #903 
Closes #904 
Closes #905


jackr276:
Memory check is clean:
image

jackr276:
The only difference is the number of lines processed. There are no actual differences in the generated code. This enhancement is going to be marked as complete and we are good for merge.

PR #908, opened 2026-05-16
[Type System]: Ternary mutability with pointers

### LINKED ISSUE #907: [Front End]: Mutability with ternaries needs to be enhanced
[Front End]: Mutability with ternaries needs to be enhanced - for ternary assignability, we always need to take the lowest mutability level
of the two values. For example, if we have mut i32* = i32* else mut i32*, this should actually fail because of the mutability levels being mismatched.

jackr276:
[Type System]: Ternary mutability with pointers - ternary mutability with pointers now works correctly. If we have pointers
on either side of the ternary, we will always take the lowest mutability level of them both.

Closes #907

### DIFF
--- a/oc/compiler/parser/parser.c
+++ b/oc/compiler/parser/parser.c
@@ -9586,8 +9586,14 @@ static generic_ast_node_t* return_statement(ollie_token_stream_t* token_stream){
 
 	//If the current function's return type is not compatible with the return type here, we'll bail out
 	if(final_type == NULL){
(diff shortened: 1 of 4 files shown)


jackr276:
Memory check is clean:
image

jackr276:
Validated:
invalid_ternary_assign_immut_to_mut_ptr.ol
ternary_assign_pointer_valid.ol (validated via run)

jackr276:
This all looks good, we are merging it up

PR #921, opened 2026-05-20
[Language Feature]: Comprehensive testing and handling for -0 values

### LINKED ISSUE #920: [Language Feature]: Ollie handling of negative 0 needs to be up-to-spec
[Language Feature]: Ollie handling of negative 0 needs to be up-to-spec - for Ollie, -0 is just 0 because we use two's complement. It should still be allowed to go through and compile though. We also need to be sure that -0.0 and -0.0d are actually negative with the sign bit set to 1.

jackr276:
[Language Feature]: Comprehensive testing and handling for -0 values

Closes #920

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -224,6 +224,21 @@ static inline u_int8_t is_f32_negative(float value){
 }
 
 
+/**
+ * Is the given f64 negative? We need to use bit manipulation to deterine
+ * this because regular float equality will not detect cases like -0.0 == 0.0
+ */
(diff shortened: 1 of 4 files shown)


jackr276:
Memory check is coming up clean:
image

jackr276:
No existing file output has changed. All new files have been configured to run via OUNIT.

jackr276:
Validated(these are all in OUNIT as well):
double_negative_0.ol
float_negative_0.ol
integer_negative_0.ol

jackr276:
This is looking good, set to merge

PR #937, opened 2026-05-26
[Instruction Selector]: Standardization of destination/operand types for binary operations

### LINKED ISSUE #938: [OIR]: Update/standardize float arithmetic operation printing
[OIR]: Update/standardize float arithmetic operation printing - they should be organized in the same way that general purpose are

jackr276:
[Instruction Selector]: Standardization of destination/operand types for binary operations

This was going to be a larger enhancement but was scrapped. There is a lot more underlying work to be done before the original end goal can be reached. However, I do think that there are some good enhancements in here that should be merged into main so we are doing it now.

Closes #938


jackr276:
Memory check is clean:
image

jackr276:
We are ready to start reviewing this. No existing files have changed, only the new testers are different.

jackr276:
Validated(these have all also been validated via run):
converting_load_in_add.ol
converting_load_in_sub.ol
load_in_add.ol
load_in_sub.ol

jackr276:
All changes are looking good, we are ready to merge.

PR #963, opened 2026-06-07
[Front End]: Refactor pointer arithmetic to occur in the CFG instead of the parser

### LINKED ISSUE #960: [Front End]: All pointer arithmetic should be handled in the CFG
[Front End]: All pointer arithmetic should be handled in the CFG - currently we have pointer arithmetic generated in the parser sometimes. This is fine but it causes us to generate instructions that could have *easily* been made into lea instructions if it was in the CFG. Proposed fix is to shift all pointer arithmetic currently in the parser into the CFG.

### LINKED ISSUE #961: [Instruction Simplifier] - Remove useless simplification logic after pointer arithmetic movement
[Instruction Simplifier] - Remove useless simplification logic after pointer arithmetic movement: we should remove this specifically marked place once we're done to improve efficiency. Booking this generic issue for tracking

### LINKED ISSUE #962: [Parser]: constant coercion for pointer arithmetic causes breakage
[Parser]: constant coercion for pointer arithmetic causes breakage - this is currently causing issues because we are attempting to force constants into pointers which is invalid. We should instead force them into an i64 when something like this happens.

### LINKED ISSUE #964: [Instruction Selector]: Memory address remediator now needs to handle leas
[Instruction Selector]: Memory address remediator now needs to handle leas - as part of the most recent enhancement this is now going to be needed

jackr276:
[Front End]: Refactor pointer arithmetic to occur in the CFG instead of the parser

Closes #960 
Closes #961 
Closes #962 
Closes #964


jackr276:
Memory check is clean:
image

jackr276:
No existing files have changed, the only difference is with the one new test file.

jackr276:
Validated:
memory_address_addition.ol

jackr276:
We are looking good to merge here

PR #1011, opened 2026-06-21
[Front End]: Ollie if-statements now branch so that the hot path in OIR is the direct fall-through path

### LINKED ISSUE #1010: [Front End]: Ollie branch remediation - we need to make sure that we fall through to hot paths
[Front End]: Ollie branch remediation - we need to make sure that we fall through to hot paths

This is a known problem with if's. We should be jumping to the "else" in an if-else chain and treat the body inside of the if as the most likely option. We need to check everything else as well to make sure that the hot path when branching is the fall-through path.

### LINKED ISSUE #1012: [Front End]: new heuristics so that we can determine if a branch is likely to execute or not
[Front End]: new heuristics so that we can determine if a branch is likely to execute or not. This is true if we have an if that does a kind of early return and/or raise error situation, or breaking out of a loop

jackr276:
[Front End]: Ollie if-statements now branch so that the hot path in OIR is the direct fall-through path

Closes #1010 
Closes #1012

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -8457,18 +8457,51 @@ static inline u_int8_t is_if_path_likely_to_execute(cfg_result_package_t* result
 	//Final block is where we need to look
 	basic_block_t* end_block = results->final_block;
 
+	//Most of our profiling is going to happen in the exit statement
+	instruction_t* exit_statement = end_block->exit_statement;
+
 	/**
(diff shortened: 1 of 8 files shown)


jackr276:
Validations were done faster than usual but this looks good. We've expanded OUNIT coverage and even added some more branch prediction capabilities. I suspect that this will come up more in the future but for now we are good.

jackr276:
We are good to merge here.

PR #1033, opened 2026-06-28
[Optimizer and Instruction Selector]: if assignment to conditional movement translation optimization

### LINKED ISSUE #1020: [OIR]: ability to emit conditional moves in OIR and translate them over to x86
[OIR]: ability to emit conditional moves in OIR and translate them over to x86 - remember that conditional moves only work for word, long word and quad word sized operands

### LINKED ISSUE #1021: [Optimizer]: ability to recognize instances where we are able to optimize into converting moves
[Optimizer]: ability to recognize instances where we are able to optimize into converting moves. This will be an optimizer pass

### LINKED ISSUE #1022: [Optimizer]: ensure that no switch is ever considered eligible for branch folding
[Optimizer]: ensure that no switch is ever considered eligible for branch folding - due to the way that switches work, it would actually decrease performance if we made them into conditional moves. For this reason we need to ensure that we never accidentally turn a switch into a conditional move chain. Testing is needed to verify this

### LINKED ISSUE #1030: [Optimizer]: verify that if-else assignments without an else are not considered
[Optimizer]: verify that if-else assignments without an else are not considered - for the way that we do this, we *must* have an else statement. As such, testing is needed to verify that if-else assignments without one are not considered.

### LINKED ISSUE #1031: [Optimizer]: Implementation of the if to conditional move converter
[Optimizer]: Implementation of the if to conditional move converter

### LINKED ISSUE #1032: [Optimizer]: if assignment moves that rely on floating point comparisons must be excluded
[Optimizer]: if assignment moves that rely on floating point comparisons must be excluded - this has to be done because floating point comparison branches may result in more than one conditional jump, which is not something that we're able to handle. This may be something that we do in the future but not right now.

### LINKED ISSUE #1034: [OIR]: Ability to have conditional moves with constant operands in the else category
[OIR]: Ability to have conditional moves with constant operands in the else category - due to the way that the translation will work, this is valid to be in the else category.

### LINKED ISSUE #1035: [Instruction Selector]: testing is needed for when we have conditional moves with byte long destinations
[Instruction Selector]: testing is needed for when we have conditional moves with byte long destinations - this is due to how byte destinations are not supported by the cmovX instruction in x86. We need testing and handling.

### LINKED ISSUE #1036: [Instruction Selector]: implementation of conditional moves with floating point branching
[Instruction Selector]: implementation of conditional moves with floating point branching - this requires extra steps using the parity flag due to NaN checking

jackr276:
[Optimizer and Instruction Selector]: if assignment to conditional movement translation optimization

The ollie compiler has been enhanced to detect and optimize eligible if-assignment statements into converting moves, eliminating the need for some branching

Closes #1020 
Closes #1021 
Closes #1022 
Closes #1030
Closes #1031 
Closes #1032 
Closes #1034 
Closes #1035 
Closes #1036


jackr276:
This is looking functional - we are ready to begin validations

jackr276:
Validated:
binary_operation_ternaries.ol
float_logical_and.ol
float_logical_and_or_mix.ol
float_logical_and_with_int.ol 
float_logical_branch.ol 
float_logical_not.ol
float_logical_not_branch.ol
float_logical_not_expanding.ol
float_logical_or.ol
float_logical_or_result_used.ol
float_multiple_logical_ands.ol
float_multiple_logical_ors.ol
function_pointer_as_parameter.ol
if-else-if-assignment-no-else.ol
if_assignment.ol
if_assignment_byte_destinations.ol
if_assignment_complex.ol 
if_assignment_converting.ol
if_assignment_float_conditional.ol
if_assignment_float_conditional.ol(KNOWN BUG TO BE ADDRESSED)
if_assignment_float_conditional_le.ol
if_assignment_not_optimizable.ol
if_assignment_variables.ol
if_else_if_assignment.ol
mutable_to_non_mutable_ternary_pointers.ol 
short_circuit_with_ternary.ol 
switch_assignment.ol
switch_ternary_input.ol
ternary_assign_pointer_valid.ol
ternary_in_for.ol
ternary_in_ternary.ol
ternary_test.ol
true_false_constants.ol

jackr276:
Looking good with the exception of one minor bug that we need to deal with, but that will be dealt with later on down the road.

PR #1074, opened 2026-07-13
[Front End]: Ollie in statements with a contiguous range are automatically converted to an efficient conditional move chain

### LINKED ISSUE #1071: [CFG]: exhaustive in optimization - if an in is exhaustive and switch eligible, we would not even need a jump table
[CFG]: exhaustive in optimization - if an in is exhaustive and switch eligible, we would not even need a jump table - since all jump targets are the same for a continuous range of values, we could just convert this into an if-else check that is true if we're within the range and false otherwise

### LINKED ISSUE #1072: [Parser]: average distance calculation is slightly off for switch, in statements
[Parser]: average distance calculation is slightly off for switch, in statements - the average distance calculation should be sum / (index - 1), because we have n - 1 differences for n numbers. This is not a huge bug but should be fixed asap.

### LINKED ISSUE #1073: [Parser & Type System]: enum types should be flagged as continuous on creation
[Parser & Type System]: enum types should be flagged as continuous on creation - instead of having to redetermine this every time for an exhaustive switch eligibility check, we should compute this once upon enum creation and then store it as a flag. This is much more efficient than what we do currently.

jackr276:
[Front End]: Ollie in statements with a contiguous range are automatically converted to an efficient conditional move chain

Closes #1071 
Closes #1072 
Closes #1073

### DIFF
--- a/oc/compiler/ast/ast.h
+++ b/oc/compiler/ast/ast.h
@@ -169,8 +169,16 @@ struct generic_ast_node_t{
 	u_int8_t dereference_needed;
 	//Is the given statement(switch or in) eligible to internally be a true switch?
 	u_int8_t is_switch_eligible;
-	//Is the given switch/in statement "exhaustive" - this is most usually false
-	u_int8_t is_exhaustive;
+	/**
+	 * This union has two values:
+	 *
+	 * 1.) Is the given switch statement "exhaustive" - this is most usually false
+	 * 2.) Is the given in statement "contiguous" - no gaps between values
+	 */
+	union {
+		u_int8_t is_exhaustive_switch;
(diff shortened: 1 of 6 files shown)


jackr276:
This is looking good. We are ready to begin validations.

jackr276:
Validated:
barely_not_contiguous_in.ol 
contiguous_in_unsigned_comparator.ol
exhaustive_in.ol
in_statement_enums.ol
in_statement_exhaustive_switch.ol
in_statement_switch_eligible_float_result.ol

jackr276:
All changed outputs look correct and OUNIT is showing no errors. We are good to merge.

PR #1077, opened 2026-07-14
[Build System]: Ollie build system initially created and tested

### LINKED ISSUE #1075: [Build System]: preprocessor integration for #link statements
[Build System]: preprocessor integration for #link statements - we need this integration for our eventual build system in Ollie that we are making

### LINKED ISSUE #1076: [Build System]: explore the idea of OC pre-linking everything into one giant token list before we go through and compile
[Build System]: explore the idea of OC pre-linking everything into one giant token list before we go through and compile - this would be a way that we could do things instead of incremental builds

### LINKED ISSUE #1078: [Build System]: change the $using directive into a plain $import
[Build System]: change the $using directive into a plain $import - the import keyword makes more sense to me intuitively

### LINKED ISSUE #1079: [Build System]: the main module may never be contained within a module
[Build System]: the main module may never be contained within a module - this would make no sense because you cannot import the main file around. The entry point must be in a "raw" file

### LINKED ISSUE #1080: [Build System]: create a separate dependency graph for our use cases
[Build System]: create a separate dependency graph for our use cases - on second thought I do think it's better if this and all its utilities are kept separate from the actual build system logic to not muddy things

### LINKED ISSUE #1081: [Symtab]: creation of a new module symtab
[Symtab]: creation of a new module symtab - in order to avoid searching more than we need to, we will maintain a symtab of all previously searched & found modules and their corresponding dependency graph nodes/token streams. This will be an additional symtab who's use will be limited to the build system. This enhancement will be added in the existing symtab file.

### LINKED ISSUE #1082: [Build System]: ability to parse import statements in the build system and search for files
[Build System]: ability to parse import statements in the build system and search for files

### LINKED ISSUE #1083: [Lexer]: we need a way to tokenize only the first 3 tokens when searching for modules
[Lexer]: we need a way to tokenize only the first 3 tokens when searching for modules

Since module searches only care about the module name, and the module name is only the first 3 tokens, when we are searching for files we should not be tokenizing the whole thing to search. Instead, we will only be tokenizing the first 3 tokens and using that to see if we have a module or not.

### LINKED ISSUE #1084: [Build System]: test required for recursive file searching inside of directories
[Build System]: test required for recursive file searching inside of directories - we need an extensive test - directories nested within directories that should be a success case. Likely that we'll need to revamp how testing works to support this but that's another discussion - this is just a basic test

### LINKED ISSUE #1085: [Lexer]: refactor the get_first_2_tokens to use the same logic as generate_all_tokens() to avoid code duplication
[Lexer]: refactor the get_first_2_tokens to use the same logic as generate_all_tokens() to avoid code duplication

### LINKED ISSUE #1086: [Build System]: local dependency searcher should only search under the root directory of where the main file(passed in via -f) resides
[Build System]: local dependency searcher should only search under the root directory of where the main file(passed in via -f) resides - we do not search above this ever if the user uses import with "<module>"

### LINKED ISSUE #1090: [Build System]: circular dependency detection is needed after imports are all created
[Build System]: circular dependency detection is needed after imports are all created

### LINKED ISSUE #1091: [Front End]: rework the entire parser flow around dependency graph nodes and a build order
[Front End]: rework the entire parser flow around dependency graph nodes and a build order - this is likely going to require big changes so booking this here to track the effort

### LINKED ISSUE #1092: [Build System]: make the circular file system detector and dependency order builder happen entirely in the existing process
[Build System]: make the circular file system detector and dependency order builder happen entirely in the existing process - we should not need an extra check for this. Since we already go in a recursive order, this should naturally be added into the current traversal. Extensive documentation is going to be need to explain how this all works.

### LINKED ISSUE #1093: [Build System]: testing is needed for third/fourth order dependency searches
[Build System]: testing is needed for third/fourth order dependency searches. We need demonstrated examples of this working and also failing to assess how well these error messages are going to be working for us.

### LINKED ISSUE #1094: [Preprocessor]: Refactor the entire preprocessor flow to use the build order instead of one token stream
[Preprocessor]: Refactor the entire preprocessor flow to use the build order instead of one token stream

### LINKED ISSUE #1095: [Test Suites]: incorporate multifile examples into the ollie run validator
[Test Suites]: incorporate multifile examples into the ollie run validator - there should be a separate pass that does this in the ollie run validator separate from the regular file pass.

### LINKED ISSUE #1096: [Ollie Run Validator]: we need a way to construct unique output .test files
[Ollie Run Validator]: we need a way to construct unique output .test files - one idea is to take every '/' and replace it with an '_', so for example ./oc/test_files/my_test.ol would have an output of oc_test_files_my_test.test. This guarantees output file uniqueness even when multifile tests are thrown into the mix.

### LINKED ISSUE #1097: [Memory Checker]: incorporate multifile test programs into the memory checker
[Memory Checker]: incorporate multifile test programs into the memory checker - similar to how we did this with the Ollie run validator, I want the memory checker to cover these as well.

jackr276:
[Build System]: Ollie build system initially created and tested

Closes #1075 
Closes #1076
Closes #1078 
Closes #1079 
Closes #1080 
Closes #1081 
Closes #1082 
Closes #1083 
Closes #1084 
Closes #1085 
Closes #1086 
Closes #1090 
Closes #1091 
Closes #1092 
Closes #1093 
Closes #1094 
Closes #1095 
Closes #1096 
Closes #1097


jackr276:
These will all need to be codified as issues eventually, but here are the general ideas after some work on this:
- we may only ever have one module per file for simplicity. The module declaration must always be the top token in the file. If we see a module declaration later down in the file, it is incorrect and will cause a complete compilation failure
- The main method may never be contained inside of a module
- module naming should be intelligent - <module_name> should be the shared library like in C, "module_name" should be the local project
- We need a way of creating /usr/lib/ollie upon installation
- We need a way to parse modules, remember there are $module directive that tell us where to wrap the module up in
- When we do parse modules, we will store their independent token streams inside of a dependency graph node. We will read token streams in one after another in a topoligical sorted fashion during actual parsing
- Modules need to have a symtab that they are stored in. This will avoid duplicate parsing and will also allow us to check for module name collisions

jackr276:
First 2 consumption appears to be working:
image

jackr276:
This is looking good for a first cut of the dependency system. No existing files have been changed and all new example programs are passing the memory and OUNIT checks. We are good to merge this in.

PR #1103, opened 2026-07-23
[Front End]: fixes to debug printing, file name & module display in debug printing

### LINKED ISSUE #1100: [Front End]: compiler flag to toggle the printing of "info" statements
[Front End]: compiler flag to toggle the printing of "info" statements - this has come to light as part of our enhancement with casting. We may not always want to see info level printing - we'd need something similar to -Wextra in gcc to display them

### LINKED ISSUE #1102: [Front End]: functions should carry the info on the modules/files that they came from
[Front End]: functions should carry the info on the modules/files that they came from

jackr276:
[Front End]: fixes to debug printing, file name & module display in debug printing

Closes #1100 
Closes #1102


jackr276:
This is looking good. As it was just a printing change there is no output that's changed. We are good to merge.

PR #1110, opened 2026-07-28
[Front End]: Definite assignment analysis & mutation analysis using SSA

### LINKED ISSUE #1108: [Front-End]: SSA definite assignment analysis
[Front-End]: SSA definite assignment analysis - we currently do assignment analysis very poorly by use of an "initialized" field. This is completely inaccurate and not what we should be doing. We should instead leverage SSA form in the CFG to perform definite assignment analysis

### LINKED ISSUE #1109: [Front End]: assigning a mutable memory region to a mutable function parameter will now count as mutation
[Front End]: assigning a mutable memory region to a mutable function parameter will now count as mutation

### LINKED ISSUE #1111: [Front End]: use synthetic initialization instead of new instruction types
[Front End]: use synthetic initialization instead of new instruction types - this will keep the IR smaller and more memory efficient, and will have the same effect for us.

### LINKED ISSUE #1118: [CFG]: all instructions now need to encode line number information
[CFG]: all instructions now need to encode line number information - this is an unfortunate side effect of definite assignment analysis. For the CFG only we'll need instructions to store the line numbers where they came from

### LINKED ISSUE #1119: [CFG]: memory region synthetic initialization
[CFG]: memory region synthetic initialization - we need to look at this as it's causing some big issues currently. My idea is that if a variable is a memory region we need to emit some kind of dummy instruction for it that symbolically represents it's initialization. This will stop the checker from hitting false positives for memory addresses.

### LINKED ISSUE #1120: [CFG]: add line number parameter to all CFG instruction emitters
[CFG]: add line number parameter to all CFG instruction emitters - this is going to suck but it needs to be done to ensure that we have line numbers everywhere

### LINKED ISSUE #1125: [CFG]: Break up the static analysis into its own separate static_analyzer module
[CFG]: Break up the static analysis into its own separate static_analyzer module

### LINKED ISSUE #1126: [Error Printing]: revamp "print_function_name" to use token array printing
[Error Printing]: revamp "print_function_name" to use token array printing - instead of generating the names we should instead be using the existing token array to just grab them from the source code itself.

### LINKED ISSUE #1127: [Error Printing]: revamp "print_variable_name" to use token array printing
[Error Printing]: revamp "print_variable_name" to use token array printing - instead of generating the names we should instead be using the existing token array to just grab them from the source code itself.

jackr276:
[Front End]: Definite assignment analysis & mutation analysis using SSA

Closes #1108 
Closes #1109 
Closes #1111 
Closes #1118 
Closes #1119 
Closes #1120
Closes #1125 
Closes #1126
Closes #1127


jackr276:
The new test files look good. This is such a large change that we're going to have to mostly trust OUNIT on this in order to review this in a reasonable amount of time. Almost all other changes are just in the error output.

jackr276:
Overall the new output looks way better. There is more work to be done with initialization states, especially with variables that have been declared as immutable and are then initialized later on down the road. I put an example below. We'll need to do more with this tomorrow, for now this is good and we are merging. 

```c
declare x:i32;

if (y == 3){
   x = 3;
} else {
 x = 4;
}
```

PR #1115, opened 2026-07-30
[Type System]: fixes to get_operand_type and rework of type widening

### LINKED ISSUE #1113: [Type System]: get operand type for relation operator does not handle signedness coercion
[Type System]: get operand type for relation operator does not handle signedness coercion - we should be seeing unsigned forcing for a comparison

### LINKED ISSUE #1114: [Type System]: basic_type_widening_coercion needs to be reworked
[Type System]: basic_type_widening_coercion needs to be reworked - currently this is just taking whichever type is bigger. This will violate what we need for signedness down the road. Instead of doing this we should instead be widening relative to the type and original size(i.e., i8 widened by 4 is i32, etc.).

### LINKED ISSUE #1116: [Front End]: relational operations between signed and unsigned can generate warnings
[Front End]: relational operations between signed and unsigned can generate warnings - like many compilers we will generate warnings about the unexpected behavior that this may cause

jackr276:
[Type System]: fixes to get_operand_type and rework of type widening. Type widening no longer requires us to have already done the signedness and floating point coercions first.

Closes #1113 
Closes #1114 
Closes #1116

### DIFF
--- a/oc/compiler/cfg/cfg.c
+++ b/oc/compiler/cfg/cfg.c
@@ -6089,7 +6089,7 @@ static inline cfg_result_package_t lower_in_expression_to_conditional_move_chain
 		add_statement(current_block, comparison_instruction);
 
 		//Get the operand type base don the two types provided
-		operand_type = get_operand_type_for_logical_operation(type_symtab, conditional_expression_variable->type, in_constant_variable->type);
+		operand_type = get_operand_type_for_relational_operation(type_symtab, conditional_expression_variable->type, in_constant_variable->type);
 
 	} else {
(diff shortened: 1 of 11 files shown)


jackr276:
This is ready to be checked. We are going to have the eventual creation of the "size" type similar to a size_t but that won't be happening for some time

jackr276:
Validated:
assign_unsigned_to_negative.ol
contiguous_in_unsigned_comparator.ol
equals_operation_signed_unsigned.ol
in_statement_enums.ol
relation_operation_unsigned.ol
short_circuit.ol

PR #1124, opened 2026-08-02
[Front End]: creation of the "size" type

### LINKED ISSUE #1121: [Front End]: support for the Ollie "size" type
[Front End]: support for the Ollie "size" type. This will work similar to "size_t" in C and will be a standardized way for us to use variable sizes.

### LINKED ISSUE #1122: [Type System]: integration of the Ollie size type into the types assignable helper
[Type System]: integration of the Ollie size type into the types assignable helper - it will behave in the same manner as a 32 bit integer

### LINKED ISSUE #1123: [Type System]: create the size type as an already aliased u32 type
[Type System]: create the size type as an already aliased u32 type - this will keep things for us a lot simpler when we are doing the internals and it makes more sense than a whole new type class

jackr276:
[Front End]: creation of the "size" type - this is really an aliased u32 but we use it for anything where a size is appropriate. It is very similar to size_t in C

Closes #1121 
Closes #1122 
Closes #1123

### DIFF
--- a/oc/compiler/ast/ast.c
+++ b/oc/compiler/ast/ast.c
@@ -57,6 +57,64 @@ void coerce_constant(generic_ast_node_t* constant_node){
 		 * type is and perform a move operation(just reassignment) to have the appropriate
 		 * expansion
 		 */
+		case SIZE_CONST:
+			switch(basic_type_token){
+				case U8:
+					constant_node->constant_type = BYTE_CONST_FORCE_U;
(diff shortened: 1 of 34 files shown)


jackr276:
All tests are working as expected with unsigned values now for the size comparison.

jackr276:
There's no need to go over tests individually here, they've been reviewed and we are good to go.

